PR #4, opened 2022-10-19
replace callback contract w. gnosis multisend contract

christopherdancy:
Description
This PR:

1. Removes the Callback contract from the repo
2. It also removes the support within the vetoguard an fractalmodule
3. The callback contract was replaced with the multisend contract since it allows us to send arb transactions and predict the gnosis safe address to remove circular dependencies.
4. The test suite goes above the call of duty in order to test the USUL contracts to ensure setup is properly completed.

Notes
Once this PR is merged into the develop branch, the updated contracts will be deployed to Goerli. Then, a PR will be created to merge develop into main.

Issue (if applicable)
Testing
To test, run:
npm test

Screenshots (if applicable)
Screenshot 2022-10-19 at 11 17 11


contracts/FractalModule.sol
```diff
@@ -22,16 +22,16 @@ contract FractalModule is IFractalModule, Module {
...
+            address _avatar,
+            address _target, 
             address[] memory _controllers // Authorized controllers
         ) = abi.decode(
                 initializeParams,
                 (address, address, address, address[])
             );
 
-        setAvatar(_avatar == address(0) ? msg.sender : _avatar);
-        setTarget(_target == address(0) ? msg.sender : _target);
```
herbig, line 34: why was this necessary previously, or was it?

herbig (approved):
awesome, so many tests 🔥🔥🔥

contracts/FractalModule.sol
christopherdancy: This was necessary to support the callback method in order to pass in the gnosis safe address. However now, it is no longer needed because we can predict the address before setting up the module.

test/Atomic-Deployment.test.ts
```diff
@@ -100,201 +139,450 @@ describe("Gnosis Safe", () => {
...
-          gnosisSingletonAddress,
-          bytecode,
-          saltNum,
-          callback.address
-        )
-      )
-        .to.emit(gnosisSafe, "ChangedGuard")
-        .withArgs(vetoGuard.address);
+    it("Setup Usul Module w/ ModuleProxyCreationEvent", async () => {
+      const VOTING_STRATEGIES_TO_DEPLOY: string[] = []; // @todo pass expected addresses for voting strategies
```
tbwebb22, line 236: Should this todo be addressed?

test/Atomic-Deployment.test.ts
```diff
@@ -100,201 +139,450 @@ describe("Gnosis Safe", () => {
...
-        [ethers.constants.AddressZero],
-        1,
-        ethers.constants.AddressZero,
-        ethers.constants.HashZero,
-        ethers.constants.AddressZero,
-        ethers.constants.AddressZero,
-        0,
-        ethers.constants.AddressZero,
+    it("Setup UsulModule w/ enabledModule event", async () => {
+      const VOTING_STRATEGIES_TO_DEPLOY: string[] = []; // @todo pass expected addresses for voting strategies
```
tbwebb22, line 359: Should this todo be addresed?

tbwebb22 (commented):
At some point, probably after we get these two existing PRs merged in, I think we should replace all the manually declared contract ABIs with actual contract typechain types, using the hardhat dependency compiler and such.

test/Atomic-Deployment.test.ts
christopherdancy: this could be removed

tbwebb22 (approved):

PR #14, opened 2022-11-29
Emit Usul metadata

### LINKED ISSUE #7: Add Proposal Metadata event to Usul Module
Usul proposals are essentally just transactions, with a transaction hash.

We would like to display additional metadata in the proposal details UI related to each, such as:

Title
Description
Anything else? @decent-dao/fractal

mudrila:
This PR implements following:

1. Implements custom `FractalUsul` contract, which extends from original `Usul`. This contract contains `submitProposalWithMetaData` which will be used instead of original `submitProposal`. 
2. It emits `ProposalMetadataCreated` with `proposalId`, `title`, `description` and `documentationUrl`
3. It stores `transactions` under `Proposal` struct during proposal creation, but also it converts `transactions` to `txHashes` and stores them as well for backwards compatibility with the rest of `Usul` code.

Closes #7

### DIFF
--- a/deploy/core/008_deploy_FractalUsul.ts
+++ b/deploy/core/008_deploy_FractalUsul.ts
@@ -11,6 +11,6 @@ const func: DeployFunction = async (hre: HardhatRuntimeEnvironment) => {
   ]);
 };
 
-func.tags = ["usul"];
+func.tags = ["FractalUsul"];
 
 export default func;


contracts/FractalUsul.sol
```diff
@@ -0,0 +1,74 @@
+//SPDX-License-Identifier: Unlicense
+pragma solidity ^0.8.0;
+
+// Usul was previously named "Seele" and SekerDAO was TokenWalk
+// that's where this naming differences are coming from
+import "@tokenwalk/seele/contracts/Usul.sol";
+
+abstract contract FractalUsul is Usul {
```
sethhrbek, line 8: what is the reasoning behind this contract being `abstract`?

contracts/FractalUsul.sol
```diff
@@ -0,0 +1,74 @@
...
+    uint256 value;
+    bytes data;
+    Enum.Operation operation;
+  }
+  
+  struct Proposal {
+    bool canceled;
+    uint256 timeLockPeriod; // queue period for safety
+    bytes32[] txHashes;
+    Transaction[] transactions;
```
sethhrbek, line 20: Are we going to store the transactions on the proposal? In order to save gas / TX costs, we discussed potentially just emitting the transactions data in an event, so that it could be retrieved off-chain later (and then submitted to execute).

contracts/FractalUsul.sol
```diff
@@ -0,0 +1,74 @@
...
+        Transaction memory currentTx = transactions[i];
+        bytes32 txHash = getTransactionHash(
+          currentTx.to, 
+          currentTx.value,
+          currentTx.data, 
+          currentTx.operation
+        );
+
+        txHashes[i] = txHash;
+      }
```
sethhrbek, line 62: Should we calculate these hashes off-chain to save gas? That would introduce a risk of the off-chain application calculating the hashes incorrectly -- but if we can trust the caller, then could be worth passing in the hashes pre-computed.

tbwebb22 (changes requested):
This PR should include some (basic) tests as well to confirm that proposals created through this new function can be successfully executed. Let me know if you want / need any help on those!

contracts/FractalUsul.sol
```diff
@@ -0,0 +1,74 @@
...
+
+  event ProposalMetadataCreated(
+    uint256 proposalId, 
+    string title, 
+    string description, 
+    string documentationUrl
+  );
+
+  // mapping of proposal id to proposal
+  mapping(uint256 => Proposal) public override proposals;
```
tbwebb22, line 33: This variable can't be overriden, but also doesn't need to be, since we can access the mapping declared in `Usul.sol`

sethhrbek (approved):

contracts/FractalUsul.sol
```diff
@@ -0,0 +1,56 @@
...
+  event ProposalMetadataCreated(
+    uint256 proposalId, 
+    Transaction[] transactions,
+    string title,
+    string description, 
+    string documentationUrl
+  );
+
+  /// @dev This method is used instead of Usul.submitProposal. Essentially - it just implements same behavior
+  /// but then - it also emits metadata of the proposal in ProposalMetadataCreated event.
```
sethhrbek, line 25: when the tests get added in, can update the natspec of this function to include the parameters.

contracts/FractalUsul.sol
```diff
@@ -0,0 +1,56 @@
...
+  function submitProposalWithMetaData(
+        Transaction[] memory transactions,
+        bytes32[] memory txHashes,
+        address strategy,
+        bytes memory data,
+        string calldata title,
+        string calldata description,
+        string calldata documentationUrl
+  ) external {
```
herbig, line 34: @decent-dao/fractal for reference, this is mostly identical to the Usul `submitProposal` function:

https://github.com/SekerDAO/Usul/blob/0cb39c0dd941b2825d401de69d16d41138a26717/contracts/Usul.sol#L216

The reason it's copied here is because that function can't be extended, and is also `external`, so can't just be called internally here

contracts/FractalUsul.sol
```diff
@@ -0,0 +1,56 @@
...
+      require(txHashes.length > 0, "proposal must contain transactions");
+
+      proposals[totalProposalCount].txHashes = txHashes;
+      proposals[totalProposalCount].strategy = strategy;
+      totalProposalCount++;
+      IStrategy(strategy).receiveProposal(
+          abi.encode(totalProposalCount - 1, txHashes, data)
+      );
+      emit ProposalCreated(strategy, totalProposalCount - 1, msg.sender);
+      emit ProposalMetadataCreated(
```
herbig, line 48: @decent-dao/fractal this is the start of what we've actually added

herbig (changes requested):
@mudrila @decent-dao/fractal I think we need to actually confirm the `txHashes` matches the transactions here.

Otherwise users can submit a list of bogus transactions to appear in the UI, along with hashes that do something totally different.

In which case we probably don't need `txHashes` as a parameter for this at all, since we'll be computing hashes from the `transactions`...

contracts/FractalUsul.sol
herbig: @mudrila I assume not since we can't override it, but is there any way we can block/revert calls to the parent `createProposal`?

contracts/FractalUsul.sol
herbig: @mudrila @christopherdancy @tbwebb22 can do do a gas test between `submitProposalWithMetaData` and `submitProposal` and see what the difference actually is?  My understanding is that events are fairly cheap, but would be good to confirm.

tbwebb22: Sure we definitely can, but the gas difference will mainly depend on how many TX's there are, and how long the three aditional metadata strings are. I'll get gas numbers for a few different scenarios

contracts/FractalUsul.sol
tbwebb22: @herbig I was thinking about this today. Unfortunately we can't prevent someone from submitting a proposal through that function, since it can't be overriden. However, we should be able to modify the UsulVetoGuard so that it doesn't allow proposals to be executed that weren't created with this new function.

contracts/FractalUsul.sol
```diff
@@ -0,0 +1,56 @@
...
+
+// Usul was previously named "Seele" and SekerDAO was TokenWalk
+// that's where this naming differences are coming from
+import "@tokenwalk/seele/contracts/Usul.sol";
+
+contract FractalUsul is Usul {
+  struct Transaction {
+    address to;
+    uint256 value;
+    bytes data;
```
herbig, line 12: @mudrila what does data this look like when it is emitted?

tbwebb22: Ah yes, so we should be able to use the Gnosis Safe API to decode it

image

contracts/FractalUsul.sol
herbig: yeah, I think that's fine, it's not worth preventing @tbwebb22 

If users approve a proposal that they don't know anything about, that's on them ha

contracts/FractalUsul.sol
```diff
@@ -0,0 +1,76 @@
...
+  ) external {
+      require(
+          isStrategyEnabled(strategy),
+          "voting strategy is not enabled for proposal"
+      );
+      require(transactions.length > 0, "proposal must contain transactions");
+
+      bytes32[] memory txHashes;
+      for (uint256 i = 0; i < transactions.length; i++) {
+        Transaction memory currentTx = transactions[i];
```
herbig, line 50: can `currentTx` be declared outside the loop and reused? it might save on gas?

contracts/FractalUsul.sol
```diff
@@ -0,0 +1,76 @@
...
+        bytes32 txHash = getTransactionHash(
```
herbig, line 51: here you could probably just do `txHashes[i] = getTransactionHash` rather than creating a `bytes32` and immediately assigning it

contracts/FractalUsul.sol
tbwebb22: Do we need to declare `currentTx` at all? Could just be 

 for (uint256 i = 0; i < transactions.length; i++) {
        bytes32 txHash = getTransactionHash(
          transactions[i].to, 
          transactions[i].value,
          transactions[i].data, 
          transactions[i].operation
        );
        txHashes[i] = txHash;
      }

herbig: Yeah I'm not sure actually, it might be that having `currentTx` declared so you don't need to reference the array location 4 times is cheaper.  But probably your suggestion is more efficient.

contracts/FractalUsul.sol
tbwebb22: Yea I like @herbig's suggestion here

contracts/FractalUsul.sol
tbwebb22: @mudrila Can you do this gas comparison when writing these tests? I can show you how I typically do it if that helps

contracts/FractalUsul.sol
```diff
@@ -0,0 +1,79 @@
...
+          transactions[i].to, 
+          transactions[i].value,
+          transactions[i].data, 
+          transactions[i].operation
+        );
+      }
+
+      proposals[totalProposalCount].txHashes = txHashes;
+      proposals[totalProposalCount].strategy = strategy;
+      totalProposalCount++;
```
tbwebb22, line 66: Small optimization, but if we move `totalProposalCount++` to the end of this function, we can remove the `- 1` in three different places

tbwebb22 (changes requested):
Looking great! Just a few suggestions

PR #17, opened 2022-12-05
Refactor guards

tbwebb22:
## Description
This PR:
1) Adds a function `getProposalTxHashes` to the `FractalUsul` contract to get all of the TX hashes associated with the specified proposal
2) Updates `UsulVetoGuard` to take advantage of this new function `getProposalTxHashes` to clean up code, and make it reference the `IFractalUsul` interface instead of `IUsul`
3) Adds the `executionPeriod` parameter to the `VetoGuard`, this gives a time limit for a TX to be executed after the TX has completed its timelock
4) Updates all the tests for these new changes
5) Deploy contracts to Goerli
6) Update NPM version

## Testing
Run `npm test`

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -20,8 +20,6 @@ If there is an accompanying Notion document, please link that here as well.
 If your testing steps are technical, outline steps for an engineer to verify.
 
 If your testing steps are functional, please provide a full guide with any features requiring special attention for operations to test your branch in the preview environment.
-
-New feature work should include a correlating automated integration test via Playwright, in collaboration with the QA team. See this repo's README.md for Playwright setup.
 ------------------------------------------------------------------------------>
 
 ## Screenshots (if applicable)


contracts/VetoGuard.sol
```diff
@@ -83,8 +91,11 @@ contract VetoGuard is
         );
 
         require(
-            transactionQueuedBlock[transactionHash] == 0,
-            "Transaction has already been queued"
+            block.timestamp >=
+                transactionQueuedTimestamp[transactionHash] +
+                    timelockPeriod +
+                    executionPeriod,
+            "Transaction has already been queued recently"
```
herbig, line 98: why was "recently" added to the end here?

herbig (approved):
This looks good to me. Would be good to have @mudrila and @christopherdancy check out as well!

contracts/VetoGuard.sol
tbwebb22: I added that because the logic there changed - The conditional that was removed there prevented the same tx from ever getting queued twice. However it should be possible to queue up a tx again if it has surpassed its execution period. This is what that new conditional allows for

herbig: ah ok, that makes sense

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,27 @@
...
+
+If there is an accompanying Notion document, please link that here as well.
+------------------------------------------------------------------------------>
+
+## Testing
+
+<!-----------------------------------------------------------------------------
+If your testing steps are technical, outline steps for an engineer to verify.
+
+If your testing steps are functional, please provide a full guide with any features requiring special attention for operations to test your branch in the preview environment.
```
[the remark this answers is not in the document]
tbwebb22: Good call, just removed this

PR #20, opened 2022-12-12
General cleanup

tbwebb22:
## Description
This PR:
1) Updates the `VetoVoting` contracts to use seconds instead of blocks for freeze period and freeze proposal period parameters to be consistent with the other contracts.
2) Adds a new test file for testing when veto / freeze when the parent DAO is a a multisig and the child DAO is Usul.
3) Deploys new contracts, increments NPM version

Note that this NPM version has not been published yet, will do so once this is approved and merged.

## Testing
Run `npm test`


herbig:
Makes sense, looks good to go!

PR #22, opened 2023-01-09
Declare subDAO via FractalRegistry

### LINKED ISSUE #21: Add DAO relationship function to FractalNameRegistry
We'd like to add a function which logs a parent/child DAO relationship to the FractalNameRegistry contract, in order to view that relationship in the DAO Hierarchy page without the need to add the parent as a signer on the child.

Spec
1. a new method on `FractalNameRegistry` which logs this relationship, similar to the naming function (e.g. `declareChildDAO(string memory _childAddress)`.
2. Rename `FractalNameRegistry` to something more applicable now that is has more utility than just naming, e.g. `FractalRegistry` / `FractalMetadataRegistry`.

https://github.com/decent-dao/fractal-contracts/blob/develop/contracts/FractalNameRegistry.sol

mudrila:
## Description

This PR implements declaring `subDAO` for the `parentDAO` the similar way as declaring DAO name - via event.
This PR does not implement any actual storing of relationships as it seems debatable and not decided completely.

## Notes

I started writing tests for atomic sub DAO deployment including step of declaring created `subDAO` under `parentDAO`, but that turned out quite a long story - so I think we should create new issue for that, as it goes way further than scope of current issue.

Closes #21 

## Testing

1. Run tests

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@fractal-framework/fractal-contracts",
-  "version": "0.1.13",
+  "version": "0.1.14",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@fractal-framework/fractal-contracts",
-      "version": "0.1.13",
+      "version": "0.1.14",
       "license": "MIT",
       "devDependencies": {
         "@gnosis.pm/safe-contracts": "^1.3.0",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@fractal-framework/fractal-contracts",
-  "version": "0.1.13",
+  "version": "0.1.14",
   "files": [
     "dist",
     "deployments",


contracts/FractalRegistry.sol
```diff
@@ -0,0 +1,20 @@
+//SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import "./interfaces/IFractalRegistry.sol";
+
+/// @notice A contract for registering Fractal DAO name strings
+/// @notice These names are non-unique, and should not be used as the identifer of a DAO
```
herbig, line 7: These comments only apply to the `updateDAOName` function, so should be moved down there

herbig: and maybe a not about what this contract does here now

herbig (approved):
Left a note about fixing a comment, but otherwise looks good to go!

tbwebb22 (changes requested):
Nice, this looks great. Only change I'll request is to increment the version in Package.json, and `npm install`

Thanks!

tbwebb22 (approved):
LGTM!

PR #31, opened 2023-02-14
Fractal usul

tbwebb22:
## Description
This PR:
1) Adds contracts from the Gnosis Zodiac Usul repo into this one. Only the contracts related to the Usul.sol and OZLinearVoting.sol (renamed to TokenLinearVoting.sol) are imported
2) Updates function names, variable names, and event names to align with Fractal conventions
3) Renames "Queue" to "Timelock" anywhere it appears in the contracts
4) Adds support for Voting implementation contracts to define which user's have the ability to create proposals
5) Update tests for all changes, add new test file for testing the FractalUsul & LinearTokenVoting contracts

## Notes
Deploying new contracts & updating NPM package will be done in a separate PR

https://github.com/decent-dao/fractal-contracts/issues/27

## Testing
1) Run `npm install`
2) Run `npm test`


contracts/FractalUsul.sol
```diff
@@ -1,86 +1,469 @@
-//SPDX-License-Identifier: MIT
+// SPDX-License-Identifier: LGPL-3.0-only
```
herbig, line 1: "LGPL imposes no conditions on using the code in software that's sold commercially."  So I believe we're good on that front

herbig: not that we have plans to, though ha

contracts/FractalUsul.sol
```diff
@@ -1,86 +1,469 @@
...
-  /// @param description - proposal description, emitted in ProposalMetadataCreated
-  /// @param documentationUrl - proposal documentation/discussion URL, emitted in ProposalMetadataCreated. 
-  /// Supposed to be link to Discord/Slack/Whatever chat discussion
-  function submitProposalWithMetaData(
+import "@gnosis.pm/zodiac/contracts/core/Module.sol";
+import "./usul/IBaseStrategy.sol";
+import "./interfaces/IFractalUsul.sol";
+
+/// @title FractalUsul - A Zodiac module that enables a voting agnostic proposal mechanism
+contract FractalUsul is Module, IFractalUsul {
```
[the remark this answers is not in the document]
tbwebb22: Yeah agreed, sounds like the group is at consensus we should rename from Usul

contracts/FractalUsul.sol
```diff
@@ -1,86 +1,469 @@
...
+    ) external {
+        require(
+            targets.length != 0,
+            "No transactions to execute provided"
+        );
+        require(
+            targets.length == values.length &&
+                targets.length == data.length &&
+                targets.length == operations.length,
+            "Array length mismatch"
```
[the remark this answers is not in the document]
tbwebb22: Hmm wouldn't that be implied? Since those four variables are the arrays that were passed in as args?

contracts/usul/LinearTokenVoting.sol
```diff
@@ -0,0 +1,138 @@
...
+        return
+            governanceToken.getPastVotes(
+                voter,
+                proposals[proposalId].startBlock
+            );
+    }
+
+    /// @notice Returns if the specified address can submit a proposal
+    /// @return bool True if the user can submit a proposal
+    function isProposer(address) public pure override returns (bool) {
```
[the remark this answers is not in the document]
tbwebb22: Based on what @herbig said, for now we just want this implementation of the voting contract to allow anyone to submit a proposal. But at least with this new function, future implementations could require some kind of token threshold to submit proposals.

@herbig Is that correct, that for now we want the LinearTokenVoting contract to allow anyone to submit a proposal?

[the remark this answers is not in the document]
herbig: yeah exactly

I don't want to change anything in our front end implementation, just allow for more complex strategies in the future

contracts/FractalUsul.sol
```diff
@@ -1,86 +1,469 @@
...
+        bytes memory initParams = abi.encode(
+            _owner,
+            _avatar,
+            _target,
+            _strategies
+        );
+        setUp(initParams);
+    }
+
+    function setUp(bytes memory initParams) public override initializer {
```
herbig, line 72: What does the initializer modifier do? Is that only callable once?

contracts/FractalUsul.sol
```diff
@@ -1,86 +1,469 @@
...
+
+    /// @notice Enables a voting strategy that can vote on proposals, only callable by the owner
+    /// @param strategy Address of the strategy to be enabled
+    function enableStrategy(address strategy) public onlyOwner {
+        require(
+            strategy != address(0) && strategy != SENTINEL_STRATEGY,
+            "Invalid strategy"
+        );
+        require(strategies[strategy] == address(0), "Strategy already enabled");
+        strategies[strategy] = strategies[SENTINEL_STRATEGY];
```
herbig, line 95: what is `SENTINEL_STRATEGY` doing exactly? Is that like the default / current strategy? I wonder if there's a better naming we can use for that

contracts/FractalUsul.sol
```diff
@@ -1,86 +1,469 @@
...
+        proposals[totalProposalCount].strategy = strategy;
+        IBaseStrategy(strategy).receiveProposal(
+            abi.encode(totalProposalCount, txHashes, data)
+        );
+        emit ProposalCreated(
+            strategy,
+            totalProposalCount,
+            msg.sender,
+            transactions,
+            title,
```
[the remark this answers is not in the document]
herbig: @tbwebb22 is it worth changing the contract to take a "metadata" string now, instead of three strings

then the metadata string we give it can be a JSON string with title/description/url instead and the frontend gets that data from the event

in a future update we can send an IPFS hash for the metadata, and achieve what you're saying without contract changes?

contracts/FractalUsul.sol
```diff
@@ -1,86 +1,469 @@
...
+            description,
+            documentationUrl
+        );
+
+        totalProposalCount++;
+    }
+
+    /// @notice Cancels an array of proposals
+    /// @param proposalIds Array of proposals to cancel
+    function cancelProposals(uint256[] memory proposalIds) external onlyOwner {
```
herbig, line 172: is this function even necessary?

I'm trying to think of a situation where a DAO would create a proposal to cancel another proposal, would that even work?

herbig: it seems like you can just vote it down instead

contracts/FractalUsul.sol
```diff
@@ -1,86 +1,469 @@
...
+            proposals[proposalId].executionCounter + targets.length
+        );
+    }
+
+    /// @notice Enables the specified array of strategy contract addresses
+    /// @param _strategies The array of strategy contract addresses
+    function setupStrategies(address[] memory _strategies) internal {
+        require(
+            strategies[SENTINEL_STRATEGY] == address(0),
+            "setupStrategies has already been called"
```
herbig, line 299: does this mean the available strategies can only be set up once?

contracts/interfaces/IFractalUsul.sol
```diff
@@ -2,44 +2,166 @@
...
-        external
-        view
-        returns (bytes32);
+    enum ProposalState {
+        Active,
+        Canceled,
+        Timelocked,
+        Executed,
+        Executable,
+        Uninitialized
```
herbig, line 14: Enum values are generally written as all caps, we could do that here now that we're cleaning it up

contracts/interfaces/IFractalUsul.sol
```diff
@@ -2,44 +2,166 @@
...
-    /// @dev Returns the hash of a transaction in a proposal.
-    /// @param proposalId the proposal to inspect.
-    /// @param index the transaction to inspect.
-    /// @return transaction hash.
-    function getTxHash(uint256 proposalId, uint256 index)
-        external
-        view
-        returns (bytes32);
+    enum ProposalState {
+        Active,
```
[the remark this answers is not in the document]
herbig: @mudrila what was the issue on the frontend?

I don't think there's a way to give them string values in solidity, but maybe something else we can do

contracts/usul/BaseTokenVoting.sol
```diff
@@ -0,0 +1,172 @@
...
+        } else {
+            revert("Invalid value for enum VoteType");
+        }
+
+        emit Voted(_voter, _proposalId, _support, _weight);
+    }
+
+    /// @notice Notifies the strategy of a new proposal, only callable by the Usul Module
+    /// @param _data Encoded proposal data, in this implementation only includes proposal ID
+    function receiveProposal(
```
herbig, line 124: I wonder if there's a better naming than "receiveProposal" for this.  To me it seems similar to the "getProposal" function name, but does something else entirely

contracts/FractalUsul.sol
tbwebb22: Correct, it makes the function only callable once

contracts/FractalUsul.sol
tbwebb22: Yeah so this is an interesting pattern, its basically a linked list that is used to track the strategy contracts. I think of the `SENTINEL_STRATEGY` as the pointer in the linked list. This is the same pattern Gnosis Safe uses for modules, with the same naming you can see here: https://github.com/safe-global/safe-contracts/blob/main/contracts/base/ModuleManager.sol

So this naming does follow Gnosis convention, but I agree its a bit confusing and am alright with renaming it if we'd like to

contracts/usul/BaseTokenVoting.sol
tbwebb22: Yeah agreed, don't really like the naming on this either. How about `initializeProposal`?

contracts/FractalUsul.sol
herbig: cool cool

contracts/FractalUsul.sol
herbig: ah, that's cool, yeah I didn't realize it was a known pattern, so let's keep the naming

contracts/FractalUsul.sol
herbig: how does it work if there is more than one strategy enabled? You can submit a proposal to either strategy?

contracts/FractalUsul.sol
herbig: hmm, I hadn't realized that.  So you can't add a voting strategy to an existing Usul module that has already been set up with one?

I suppose that's not a huge issue currently, but I wonder if there's a way we can allow for strategies to be added after the fact

I could see a DAO starting as a linear token voting DAO, but adding quadratic voting later one.

But maybe in that case they would just have to remove the current module and add a new one with the proper strategy.

contracts/usul/BaseTokenVoting.sol
herbig: yeah that sounds good, thank you sir

contracts/usul/BaseTokenVoting.sol
```diff
@@ -0,0 +1,172 @@
...
+import "@openzeppelin/contracts-upgradeable/utils/cryptography/draft-EIP712Upgradeable.sol";
+import "@openzeppelin/contracts/utils/cryptography/ECDSA.sol";
+import "./BaseStrategy.sol";
+
+/// @title An abstract contract used as a base for ERC-20 token voting strategies
+abstract contract BaseTokenVoting is BaseStrategy {
+    enum VoteType {
+        Against,
+        For,
+        Abstain
```
herbig, line 13: We should capitalize this one too

herbig: Also maybe we should change it to YES, NO, ABSTAIN, since they're referred to `yesVotes` / `noVotes` below, rather than for/against, or alternatively change those to `forVotes` / `againstVotes`

contracts/usul/BaseTokenVoting.sol
```diff
@@ -0,0 +1,172 @@
...
+        mapping(address => bool) hasVoted;
+    }
+
+    uint256 public votingPeriod; // the length of time voting is valid for a proposal
+    uint256 public timelockPeriod;
+    string public name;
+
+    mapping(uint256 => ProposalVoting) internal proposals;
+
+    event TimelockPeriodUpdated(uint256 newTimeLockPeriod);
```
herbig, line 31: super nitpicky but we should be consistent with making Timelock a single word, so this should be `newTimelockPeriod` (lowercase L)

contracts/usul/BaseTokenVoting.sol
```diff
@@ -0,0 +1,172 @@
...
+        address voter,
+        uint256 proposalId,
+        uint8 support,
+        uint256 weight
+    );
+
+    /// @notice Updates the voting time period
+    /// @param _newVotingPeriod The voting time period in seconds
+    function updateVotingPeriod(uint256 _newVotingPeriod) external onlyOwner {
+        _updateVotingPeriod(_newVotingPeriod);
```
herbig, line 45: what is the purpose of having these external functions call internal functions here? couldn't we just have the login in the function?

contracts/interfaces/IFractalUsul.sol
herbig, line 14: in what instance would a proposal be `Uninitialized`?  A user would need to submit a proposal when there is no strategy enabled?

I'm just wondering if there's a way we can refactor that state out and handle not getting to it with `require` checks instead

contracts/interfaces/IFractalUsul.sol
```diff
@@ -2,44 +2,166 @@
...
+        Canceled,
```
herbig, line 10: if we get rid of the cancel proposal method we should be able to get rid of this state as well, right?

contracts/interfaces/IFractalUsul.sol
```diff
@@ -2,44 +2,166 @@
...
+        Timelocked,
+        Executed,
+        Executable,
```
herbig, line 13: can you reorder this to its chronological order, before executed

contracts/interfaces/IFractalUsul.sol
```diff
@@ -2,44 +2,166 @@
...
+    ) external;
+
+    /// @notice Cancels an array of proposals
+    /// @param proposalIds Array of proposals to cancel
+    function cancelProposals(uint256[] memory proposalIds) external;
+
+    /// @notice Called by the strategy contract when the proposal vote has succeeded
+    /// @param proposalId the identifier of the proposal
+    /// @param timeLockPeriod the optional delay time
+    function timelockProposal(uint256 proposalId, uint256 timeLockPeriod) external;
```
herbig, line 49: another instance of timeLockPeriod that should be lowercase L

contracts/interfaces/IFractalUsul.sol
```diff
@@ -2,44 +2,166 @@
...
+        address[] memory targets,
+        uint256[] memory values,
+        bytes[] memory data,
+        Enum.Operation[] memory operations
+    ) external;
+
+    /// @notice Returns if a strategy is enabled
+    /// @param _strategy The address of the strategy to check
+    /// @return True if the strategy is enabled
+    function isStrategyEnabled(address _strategy) external view returns (bool);
```
herbig, line 83: we should be consistent in whether we preface parameters with an underscore, I don't have a preference either way, but it seems like this is the only one with it

contracts/interfaces/IFractalUsul.sol
```diff
@@ -2,44 +2,166 @@
...
+
+    /// @notice Returns array of strategy contract addresses
+    /// @param startAddress Address in the strategy linked list to start with
+    /// @param count Maximum number of strategies that should be returned
+    /// @return strategiesArray Array of strategy
+    /// @return next Next address in the linked list
+    function getStrategies(
+        address startAddress,
+        uint256 count
+    ) external view returns (address[] memory strategiesArray, address next);
```
herbig, line 93: since it's already an array, it feels like we can just call this `strategies`

contracts/usul/BaseStrategy.sol
```diff
@@ -0,0 +1,58 @@
...
+        require(
+            msg.sender == address(usulModule),
+            "Only callable by Usul module"
+        );
+        _;
+    }
+
+    /// @notice Sets the address of the Usul contract, only callable by owner
+    /// @param _usulModule The address of the Usul module
+    function setUsul(address _usulModule) external onlyOwner {
```
herbig, line 29: I referenced this is a previous comment, but yeah, we should be consistent with underscore / no underscore for parameters throughout the code

contracts/usul/BaseQuorumPercent.sol
```diff
@@ -0,0 +1,33 @@
+// SPDX-License-Identifier: LGPL-3.0-only
+pragma solidity ^0.8.0;
+
+import "@openzeppelin/contracts-upgradeable/access/OwnableUpgradeable.sol";
+
+/// @title BaseQuorumPercent - A Usul strategy extension that enables percent based quorums
+abstract contract BaseQuorumPercent is OwnableUpgradeable {
```
herbig, line 7: why is this `OwnableUpgradeable` instead of just `Ownable`? Is that necessary here?

contracts/usul/BaseStrategy.sol
```diff
@@ -0,0 +1,58 @@
+// SPDX-License-Identifier: LGPL-3.0-only
+pragma solidity ^0.8.0;
+
+import "../interfaces/IFractalUsul.sol";
+import "./IBaseStrategy.sol";
+import "@gnosis.pm/zodiac/contracts/factory/FactoryFriendly.sol";
+import "@openzeppelin/contracts-upgradeable/access/OwnableUpgradeable.sol";
+
+abstract contract BaseStrategy is
+    OwnableUpgradeable,
```
herbig, line 10: same with this, can this just be `Ownable`?

contracts/usul/BaseQuorumPercent.sol
```diff
@@ -0,0 +1,33 @@
...
+
+    event QuorumNumeratorUpdated(uint256 newQuorumNumerator);
+
+    function quorum(uint256 blockNumber) public view virtual returns (uint256);
+
+    function updateQuorumNumerator(
+        uint256 newQuorumNumerator
+    ) public virtual onlyOwner {
+        _updateQuorumNumerator(newQuorumNumerator);
+    }
```
herbig, line 19: another instance I'm not sure it's necessary to have `public` / `internal` versions of, it seems like we could just have a single `external` function

contracts/FractalUsul.sol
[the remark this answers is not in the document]
tbwebb22: Yeah for sure, I like that. Just a single metadata string makes sense

contracts/FractalUsul.sol
tbwebb22: Ok cool, will remove

contracts/UsulVetoGuard.sol
```diff
@@ -42,49 +42,49 @@ contract UsulVetoGuard is
...
-        proposals[proposalId].queuedBlock = block.number;
+        proposals[proposalId].timelockedBlock = block.number;
 
         bytes32[] memory txHashes = fractalUsul.getProposalTxHashes(proposalId);
         for(uint256 i; i < txHashes.length; i++) {
           transactionToProposal[txHashes[i]] = proposalId;
         }
 
-        emit ProposalQueued(msg.sender, proposalId);
+        emit ProposalTimelocked(msg.sender, proposalId);
```
sethhrbek, line 87: minor gas optimization -- cache `txHashes.length` in its own variable before the loop on line 83.
```
uint256 txHashesLength = txHashes.length;
for(uint256 i; i < txHashesLength; i++)
```

contracts/FractalUsul.sol
herbig: ah, ok cool, that's good, I think I misread how that works, definitely something we want to be able to do so that's great

sethhrbek (commented):
https://gist.github.com/sethhrbek/ffa9be7e164fe2e0e31598fcf824dcfc

This is report generated by code4arena's static analyzer. It includes list of gas optimizations. 

@herbig @tbwebb22 -- some of these may/may not be worth implementing. Any that are deemed worthy could happen in a followup task.

contracts/usul/BaseTokenVoting.sol
tbwebb22: Pushed this one up

contracts/FractalUsul.sol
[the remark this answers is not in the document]
herbig: nice, yeah be sure to add that to the list of frontend changes

contracts/FractalUsul.sol
tbwebb22: Removed this function

contracts/usul/BaseTokenVoting.sol
tbwebb22: Fixed this

contracts/usul/BaseTokenVoting.sol
tbwebb22: So it just DRYs up the code a bit. These internal functions can be used by the voting contracts that inherit from `BaseTokenVoting`. For example, `LinearTokenVoting` calls both `updateVotingPeriod` and `updateTimelockPeriod` from within the `setUp` function

contracts/interfaces/IFractalUsul.sol
tbwebb22: Yeah, it would only be `Uninitialized` when you query a proposal ID that hasn't been submitted yet. 

I agree this state isn't necessary, and should revert instead, so I just pushed a commit to move this into a require statement check, and removed that `Uninitialized` state from the enum.

contracts/interfaces/IFractalUsul.sol
tbwebb22: Yep! Just pushed commit to remove this state

contracts/interfaces/IFractalUsul.sol
tbwebb22: Done

contracts/interfaces/IFractalUsul.sol
tbwebb22: Fixed this

contracts/interfaces/IFractalUsul.sol
tbwebb22: `strategies` collides with the global state variable name, so I renamed it to `_strategies`

contracts/usul/BaseStrategy.sol
tbwebb22: Hmm well in this case, `usulModule` would collide with the global variable name. We could go with underscores on all function args tho?

contracts/usul/BaseQuorumPercent.sol
tbwebb22: So both `Owanble` and `OwnableUpgradeable` allows the `owner` to set a new `owner`.

The reason we need to use `OwnableUpgradeable` here is because this contract won't be set up using a constructor, but instead using an initializer function. `Ownable` has a constructor, which means that any contracts inheriting from it need to call that from their own constructor.

contracts/usul/BaseQuorumPercent.sol
tbwebb22: So yeah this is also just to DRY the code up. `LinearTokenVoting` calls the internal `_updateQuorumNumerator` from within its `setUp` function.

contracts/usul/BaseStrategy.sol
tbwebb22: Same as above

contracts/UsulVetoGuard.sol
tbwebb22: Hmm just A/B tested this, caching the array length used 5 more gas than the original code

contracts/interfaces/IFractalUsul.sol
tbwebb22: Yeah, agree we should be consistent here. I do prefer having the underscore on function args, since a lot of times they would otherwise collide with global state variable names. I can change them to all have underscores

contracts/usul/BaseTokenVoting.sol
tbwebb22: Yeah sounds good. I like Yes / No more than For / Against

tbwebb22: Done

contracts/interfaces/IFractalUsul.sol
herbig: yeah that makes sense, underscores actually makes sense for parameters in that case

contracts/usul/BaseStrategy.sol
herbig: yeah, going with underscores for all of them seems like a Solidity standard, so makes sense

contracts/interfaces/IFractalUsul.sol
herbig: awesome

contracts/interfaces/IFractalUsul.sol
```diff
@@ -2,44 +2,161 @@
...
-    /// @return transaction hash.
-    function getTxHash(uint256 proposalId, uint256 index)
-        external
-        view
-        returns (bytes32);
+    enum ProposalState {
+        ACTIVE,
+        TIMELOCKED,
+        EXECUTABLE,
+        EXECUTED
```
herbig, line 12: does this need a `FAILED` state of some sort?

If doesn't pass it doesn't make sense to stay in the ACTIVE state

herbig: we discussed this this morning,

it's possible some strategies wouldn't have the concept of quorum, so it's probably best to have a single "FAILED" state, and let the front end figure out why

contracts/usul/LinearTokenVoting.sol
```diff
@@ -0,0 +1,138 @@
...
+/// @title A Usul strategy that enables linear token voting
+contract LinearTokenVoting is BaseTokenVoting, BaseQuorumPercent {
+    ERC20Votes public governanceToken;
+
+    constructor(
+        address _owner,
+        ERC20Votes _governanceToken,
+        address _usulModule,
+        uint256 _votingPeriod,
+        uint256 quorumNumerator_,
```
herbig, line 17: why is this underscore at the end here?

herbig: as well as `name_` below

contracts/usul/LinearTokenVoting.sol
```diff
@@ -0,0 +1,138 @@
...
+            quorumNumerator_,
+            _timelockPeriod,
+            name_
+        );
+        setUp(initParams);
+    }
+
+    /// @notice Sets up the contract with initial parameters
+    /// @param initParams The initial setup parameters encoded as bytes
+    function setUp(bytes memory initParams) public override initializer {
```
herbig, line 35: can this be private?

contracts/FractalUsul.sol
```diff
@@ -1,86 +1,445 @@
-//SPDX-License-Identifier: MIT
```
herbig, line 1: I think it might make sense to put this into the "usul" directory with the other Usul stuff (to be renamed whatever we go with)

also maybe Usul related interfaces within a usul/interfaces directory?  I'm not sure, would just like to have the "Usul" framework a self contained directory independent of any other contracts

contracts/usul/LinearTokenVoting.sol
```diff
@@ -0,0 +1,138 @@
...
+            msg.sender,
+            _support,
+            getVotingWeight(msg.sender, _proposalId)
+        );
+    }
+
+    /// @notice Returns if a proposal has succeeded
+    /// @param _proposalId The ID of the proposal to vote for
+    /// @return bool True if the proposal has passed
+    function isPassed(uint256 _proposalId) public view override returns (bool) {
```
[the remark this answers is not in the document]
herbig: oh yeah for sure, this should just return false if the conditions aren't met, rather than reverting

contracts/usul/BaseTokenVoting.sol
```diff
@@ -0,0 +1,172 @@
...
+        proposals[proposalId].startBlock = block.number;
+
+        emit ProposalReceived(proposalId, block.timestamp);
+    }
+
+    /// @notice Calls the Usul module to notify that a quorum has been reached
+    /// @notice Timelocks the proposal and starts timelock period
+    /// @param _proposalId The ID of the proposal to timelock
+    function timelockProposal(uint256 _proposalId) public virtual override {
+        require(isPassed(_proposalId));
```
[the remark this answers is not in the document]
herbig: I saw this missing message in a PR I did for some formatting tweaks to this:

https://github.com/decent-dao/fractal-contracts/pull/32

But I can't get the tests to run locally so I can fix all the ones that broke because I tweaked all the messages..

Might need your help @tbwebb22 on Monday to get set up locally to run the tests

contracts/usul/LinearTokenVoting.sol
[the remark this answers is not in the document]
tbwebb22: Yeah agreed, this makes sense

contracts/usul/LinearTokenVoting.sol
tbwebb22: No, this needs to be able to be called by the ModuleProxyFactory, since we are using the clone pattern here.

contracts/FractalUsul.sol
tbwebb22: Yeah agreed, this makes sense too

contracts/interfaces/IFractalUsul.sol
tbwebb22: Cool, just pushed up changes for adding the FAILED state

herbig: sick

contracts/usul/LinearTokenVoting.sol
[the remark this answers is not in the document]
tbwebb22: Pushed up this change

contracts/usul/BaseTokenVoting.sol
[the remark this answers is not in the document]
tbwebb22: Yeah so I'm guessing the tests aren't running due to not having the provider set in .env.

I did push up a change that adds a comment to this require statement

contracts/usul/LinearTokenVoting.sol
tbwebb22: Good call, yeah this was leftover from Usul. Just pushed up fix for this

tbwebb22:
> 



> @tbwebb22 there's a few Usul references in comments, but other than that this is good to goooooo
> 
> Screen Shot 2023-02-21 at 10 44 32 AM

Ah good catch, just fixed these

contracts/azorius/Azorius.sol
```diff
@@ -0,0 +1,463 @@
...
+            _strategy != address(0) && _strategy != SENTINEL_STRATEGY,
+            "Invalid strategy"
+        );
+        require(
+            strategies[_strategy] == address(0),
+            "Strategy already enabled"
+        );
+
+        strategies[_strategy] = strategies[SENTINEL_STRATEGY];
+        strategies[SENTINEL_STRATEGY] = _strategy;
```
sethhrbek, line 88: In `setupStrategies` --

```
        require(
            strategies[SENTINEL_STRATEGY] == address(0),
            "setupStrategies has already been called"
        );
```

It checks to make sure that `strategies[SENTINEL_STRATEGY] == address(0)` --- but after calling `enableStrategy`, wouldn't that update `strategies[SENTINEL_STRATEGY]` to not be address 0, thus allowing `setupStrategies` to be called again?

`setupStrategies` is only called by `setUp`, so maybe the check in `setupStrategies` is only guarding against `setUp` being called multiple times?

contracts/azorius/Azorius.sol
```diff
@@ -0,0 +1,463 @@
...
+    /// @param _strategy Address of the strategy to be enabled
+    function enableStrategy(address _strategy) public onlyOwner {
+        require(
+            _strategy != address(0) && _strategy != SENTINEL_STRATEGY,
+            "Invalid strategy"
+        );
+        require(
+            strategies[_strategy] == address(0),
+            "Strategy already enabled"
+        );
```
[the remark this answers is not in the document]
herbig: @sethhrbek does not using custom errors save on gas?  If so yeah I'd vote for just cutting them all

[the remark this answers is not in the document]
sethhrbek: @herbig --

Using custom errors with revert could potentially save on gas if we didn't use a revert string.

```
error InvalidStrategy();

if (_strategy == address(0) || _strategy == SENTINEL_STRATEGY) {
  revert InvalidStrategy();
}
```

If we could also pass an argument to the custom error to get more info (`error InvalidStrategy(address strategy)`).

The front-end would need to decode the custom error selector (see "errors with parameters" here) https://blog.soliditylang.org/2021/04/21/custom-errors/. Ethers has some updates since this article was written which should make catching the errors easier. 

There would be gas savings on deploy, but runtime gas savings would only happen if a revert is hit. Gas savings may be negligible enough to not warrant changing these, but thought it was worth a discussion.

contracts/azorius/Azorius.sol
tbwebb22: Hmm so once `enableStrategy` is called, `strategies[SENTINEL_STRATEGY]` will no longer be equal to `address(0)`, so then `setupStrategies` will not be able to be called again, right?

contracts/azorius/Azorius.sol
```diff
@@ -0,0 +1,463 @@
...
+
+/// @title Azorius - A Zodiac module that enables a voting agnostic proposal mechanism
+contract Azorius is Module, IAzorius {
+    bytes32 public constant DOMAIN_SEPARATOR_TYPEHASH =
+        0x47e79534a245952e8b16893a336b85a3d9ea9fa8c573f3d803afb92a79469218;
+    bytes32 public constant TRANSACTION_TYPEHASH =
+        0x72e9670a7ee00f5fbf1049b8c38e3f22fab7e9b85029e85cf9412f17fdd5c2ad;
+    uint256 public totalProposalCount; // Total number of submitted proposals
+    uint256 public timelockPeriod; // Delay between a proposal is passed and can be executed
+    address internal constant SENTINEL_STRATEGY = address(0x1);
```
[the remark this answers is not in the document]
tbwebb22: Yeah, so this naming was leftover from Usul. This is the same pattern used by Gnosis Safe for managing module addresses: 

https://github.com/safe-global/safe-contracts/blob/main/contracts/base/ModuleManager.sol#L31

https://en.wikipedia.org/wiki/Sentinel_value

contracts/azorius/Azorius.sol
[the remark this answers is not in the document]
tbwebb22: Yeah, I'm definitely interested in implementing these custom errors for Fractal. Maybe we can include these in this next PR if we think its worth it.

contracts/AzoriusVetoGuard.sol
```diff
@@ -42,49 +42,49 @@ contract UsulVetoGuard is
 
         transferOwnership(_owner);
         vetoVoting = IVetoVoting(_vetoVoting);
-        baseStrategy = IBaseStrategy(_baseStrategy);
-        fractalUsul = IFractalUsul(_fractalUsul);
+        strategy = IBaseStrategy(_strategy);
+        fractalAzorius = IAzorius(_fractalAzorius);
         executionPeriod = _exeuctionPeriod;
```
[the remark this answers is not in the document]
tbwebb22: Oh good catch, just fixed this on the new branch. We won't be getting rid of this contract, just renaming it to `AzoriusFreezeGuard`

contracts/azorius/Azorius.sol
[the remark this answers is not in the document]
herbig: yeah, we should keep it since it follows their pattern, but some comments like @sethhrbek said is a good call.  I would even link to the wiki for it ha

contracts/azorius/Azorius.sol
sethhrbek: Oh yeah I read this wrong. Have been writing custom errors where the conditional logic is the inverse of "require" which is the cause.  All is good here, can ignore the comment.

PR #35, opened 2023-03-09
IAzorius documentation

herbig:
## Description

Fleshing out documentation for the IAzorius interface.


contracts/azorius/interfaces/IAzorius.sol
```diff
@@ -3,22 +3,65 @@ pragma solidity ^0.8.0;
...
+        // hashes of the transactions that are being proposed
+        bytes32[] txHashes;
+        // time (in seconds) this proposal will be timelocked for if it passes
+        uint256 timelockPeriod;
+        // time (in seconds) this proposal has to be executed after timelock
+        // ends before it is expired
+        uint256 executionPeriod;
+        // count of transactions that have been executed within the proposal
+        uint256 executionCounter;
     }
```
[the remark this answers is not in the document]
herbig: yeah fair, I'll update

contracts/azorius/interfaces/IAzorius.sol
```diff
@@ -28,38 +71,59 @@ interface IAzorius {
...
 
-    /// @notice Executes the specified transaction within a proposal
-    /// @notice Transactions must be called in order
-    /// @param _proposalId the identifier of the proposal
-    /// @param _target the contract to be called by the avatar
-    /// @param _value ether value to pass with the call
-    /// @param _data the data to be executed from the call
-    /// @param _operation Call or Delegatecall
+    /**
+     * Executes the specified Proposal.
```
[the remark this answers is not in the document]
herbig: yeah good call, I'll add more context around that, now that we're making it private as well

herbig:
> Looks good! Main thing I'm worried about is removing the natspec from the Azorius contract. Why would we not just want to include the documentation in both the contract and the interface?

@tbwebb22 having the same comments repeated in two locations makes it likely they'll get out of sync, or at the least require updating two places at once to change, rather than having the implementation just say go look at the interface

contracts/azorius/Azorius.sol
```diff
@@ -168,13 +158,16 @@ contract Azorius is Module, IAzorius {
...
+    /**
+     * Executes the specified transaction in a Proposal, by index.
+     * Transactions in a proposal must be called in order.
+     *
+     * @param _proposalId identifier of the proposal
+     * @param _target contract to be called by the avatar
+     * @param _value ETH value to pass with the call
+     * @param _data data to be executed from the call
+     * @param _operation Call or Delegatecall
+     */
```
sethhrbek, line 170: would a more accurate name be `executeProposalTxByIndex`?

contracts/azorius/interfaces/IAzorius.sol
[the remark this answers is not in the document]
sethhrbek: left a comment about potentially updating the name of the function as well

contracts/azorius/Azorius.sol
tbwebb22: Yeah, agreed I think that's a better description

tbwebb22:
> 

Ok, I'm cool with this. I dug into it a bit more, didn't know that the Natspec was also inherited from base functions: 
https://docs.soliditylang.org/en/v0.8.17/natspec-format.html#inheritance-notes

Also didn't know about that `@inheritdoc` tag which is good to know

contracts/azorius/Azorius.sol
tbwebb22: @herbig Do you want to address this function name change in this PR, or would you like me to change it my current branch?

herbig:
> > 
> 
> Ok, I'm cool with this. I dug into it a bit more, didn't know that the Natspec was also inherited from base functions: https://docs.soliditylang.org/en/v0.8.17/natspec-format.html#inheritance-notes
> 
> Also didn't know about that `@inheritdoc` tag which is good to know

oh cool yeah that's good to know, I should remove all the docs on the Azorius contract that are implementations of the interface, or add that tag

contracts/azorius/Azorius.sol
herbig: I can update this, no prob

tbwebb22: Oh yeah, also making this private will cause some tests to fail. I can push commit to fix those on this branch unless you want to tackle it

herbig: Ah true, yeah might be quicker for you to fix those: https://github.com/decent-dao/fractal-contracts/pull/35/commits/662162003dd69fa264b3d74204d79de8ceb52c01

tbwebb22: This is done

tbwebb22 (approved):

sethhrbek (approved):

PR #36, opened 2023-03-14
Timestamps to blocks

tbwebb22:
## Description
This PR updates the contracts to use blocks and block numbers instead of block timestamps for measuring periods of time. It also updates the tests to reflect these changes.

## Testing
Run `npm test`


contracts/BaseFreezeVoting.sol
```diff
@@ -7,10 +7,9 @@ import "./interfaces/IBaseFreezeVoting.sol";
 abstract contract BaseFreezeVoting is FactoryFriendly, IBaseFreezeVoting {
     uint256 public freezeVotesThreshold; // Number of freeze votes required to activate a freeze
     uint256 public freezeProposalCreatedBlock; // Block number the freeze proposal was created at
-    uint256 public freezeProposalCreatedTime; // Timestamp the freeze proposal was created at
```
herbig, line 10: do we not need this anymore?

herbig: oh shoot nevermind, just the block now ha, I'm dumb

herbig (approved):
Looks good. @tbwebb22 I had two commits to clean up some comments as well, just FYI

PR #40, opened 2023-03-16
Custom errors

tbwebb22:
## Description
This PR:
1) Converts all require statements into custom error messages
2) Renames "Token" to "ERC20" everywhere its relevant

## Testing
Run `npm test`


contracts/FractalModule.sol
```diff
@@ -45,10 +49,7 @@ contract FractalModule is IFractalModule, Module {
             bytes memory _data,
             Enum.Operation _operation
         ) = abi.decode(execTxData, (address, uint256, bytes, Enum.Operation));
-        require(
-            exec(_target, _value, _data, _operation),
-            "Module transaction failed"
-        );
+        if(!exec(_target, _value, _data, _operation)) revert TxFailed();
```
herbig, line 52: if the `exec` function fails here can it even continue anyway? wouldn't `exec` revert?

tbwebb22: Nope, `exec` will return `true` or `false` depending on whether the transaction fails or succeeds, but shouldn't revert. 

https://github.com/gnosis/zodiac/blob/master/contracts/core/Module.sol#L37

herbig (approved):

herbig:
Looks good, gonna merge this one in so we can move forward!

PR #42, opened 2023-03-16
Allow for proposals with 0 transactions

herbig:
## Description

Multiple beta feedback users have mentioned the desire to conduct votes on chain for things that do not necessarily involve transactions.

This change allows us to build such a feature into the frontend if we decide to do so, without adding any additional complexity to the contracts.

### DIFF
--- a/contracts/azorius/Azorius.sol
+++ b/contracts/azorius/Azorius.sol
@@ -120,10 +120,6 @@ contract Azorius is Module, IAzorius {
     ) external {
         require(isStrategyEnabled(_strategy), "Voting strategy is not enabled");
         require(
-            _transactions.length > 0,
-            "Proposal must contain at least one transaction"
-        );
-        require(
             IBaseStrategy(_strategy).isProposer(msg.sender),
             "Caller cannot submit proposals"
         );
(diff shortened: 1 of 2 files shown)


Da-Colon (commented):
Does this break anything down the workflow? How do we handle queue and execute on the frontend? would these still go as normal?

sethhrbek (commented):
In `submitProposal`, if TX count is 0, do we need to set `txHashes`, `timelockPeriod`, or `executionPeriod`? Curious if it saves gas to skip setting those.

herbig:
> In `submitProposal`, if TX count is 0, does it need to set `txHashes`, `timelockPeriod`, or `executionPeriod`? Curious if it saves gas to skip setting those.

You wouldn't need to set these, but I didn't want to get too into building a lot of logic around not having transactions, just make it possible not to have them.

But I could see not setting them at all yeah

sethhrbek (approved):

contracts/azorius/Azorius.sol
```diff
@@ -318,6 +314,11 @@ contract Azorius is Module, IAzorius {
...
         } else if (!_strategy.isPassed(_proposalId)) {
             return ProposalState.FAILED;
+        } else if (_proposal.txHashes.length == 0) {
+            // a Proposal with 0 transactions goes straight to EXECUTED
+            // this allows for the potential for on-chain voting for 
+            // "off-chain" executed decisions
+            return ProposalState.EXECUTED;
         } else if (
             block.number <= votingEndBlock + _proposal.timelockPeriod
         ) {
```
tbwebb22, line 324: I think its unnecessary to add an additional else block for this special case, if we move the original 'EXECUTED' block up above the 'TIMELOCKED' block, it will catch this special case. So could be simplified to:
image

tbwebb22 (commented):
Looks good! Just one comment on a way to clean up the `proposalState` logic

tbwebb22 (approved):
LGTM

PR #44, opened 2023-03-19
Azorius Proposal based subDAO freezing

herbig:
## Description

Refactors our Freeze voting contracts and guard in order to remove the voting aspect and allow DAOs to freeze via a normal Azorius Proposal instead.

## Testing

TODO... Will fix the tests provided there is no issue with implementing freezing in this way.


herbig:
@tbwebb22 as discussed we'll hold off on making major changes here, but something to consider later on

PR #47, opened 2023-03-23
key/value pair logger

herbig:
## Description

A suggestion from @tbwebb22 when we were discussing the DAOname was to have an arbitrary key/value pair contract.

I wanted to get this in as well, so that it is included in the audit, as we may make use of this with the Snapshot integration at some point soon.


contracts/KeyValuePair.sol
```diff
@@ -0,0 +1,15 @@
...
+
+import "./interfaces/IKeyValuePair.sol";
+
+/**
+ * A simple contract to log key/value pair events for the calling address.
+ */
+contract KeyValuePair is IKeyValuePair {
+
+    /** @inheritdoc IKeyValuePair*/
+    function updateValue(string memory _key, string memory _value) external {
```
tbwebb22, line 12: Should we make this function receive an array of keys and values? So that it could emit an arbitrary number of keys and values at once?

contracts/KeyValuePair.sol
```diff
@@ -0,0 +1,31 @@
...
+        uint256 keyCount = _keys.length;
+
+        if (keyCount != _values.length)
+            revert IncorrectValueCount();
+
+        for (uint256 i; i < keyCount; ) {
+            emit ValueUpdated(msg.sender, _keys[i], _values[i]);
+            unchecked {
+                ++i;
+            }
```
tbwebb22, line 28: nice

PR #49, opened 2023-03-25
Allow for Proposer threshold and super majority basis

herbig:
## Description

From feedback with Bancor folks, one of the things they brought up was the ability to set a super majority for voting strategies.

Currently we only support a simple majority, but this change modifies `LinearERC20Voting` to have a parameter, `basisNumerator`, which designates the basis percentage to use when calculating whether a proposal passes.

Additionally while implementing this, I decided to add the ability to set a threshold for how many tokens are required to submit proposals.

## Testing

I'll need to flesh out all of these tests to make ensure this works properly.. Opening this up for discussion.


contracts/azorius/LinearERC20Voting.sol
```diff
@@ -153,17 +171,13 @@ contract LinearERC20Voting is BaseStrategy, BaseQuorumPercent {
...
-            return true;
-        }
-
-        return false;
+        return (
+            proposalVotes[_proposalId].votingEndBlock != 0 && // end block wasn't set to 0 TODO why do we need this?
+            block.number > proposalVotes[_proposalId].votingEndBlock && // voting period has ended
+            proposalVotes[_proposalId].yesVotes >= quorum(proposalVotes[_proposalId].votingStartBlock) && // yes votes meets the quorum
+            block.number > proposalVotes[_proposalId].votingEndBlock && // more yes votes than no (simple majority)
+            proposalVotes[_proposalId].yesVotes / (proposalVotes[_proposalId].yesVotes + proposalVotes[_proposalId].noVotes) > basis() // yes votes meets the basis
```
herbig, line 179: something up with this calculation but I'm not quite sure....

contracts/azorius/LinearERC20Voting.sol
```diff
@@ -153,17 +171,13 @@ contract LinearERC20Voting is BaseStrategy, BaseQuorumPercent {
...
-            quorum(proposalVotes[_proposalId].votingStartBlock) &&
-            proposalVotes[_proposalId].votingEndBlock != 0 &&
-            block.number > proposalVotes[_proposalId].votingEndBlock
-        ) {
-            return true;
-        }
-
-        return false;
+        return (
+            proposalVotes[_proposalId].votingEndBlock != 0 && // end block wasn't set to 0 TODO why do we need this?
```
herbig, line 175: @tbwebb22 is it necessary to check `votingEndBlock != 0` here?

contracts/azorius/BaseVotingBasisPercent.sol
```diff
@@ -0,0 +1,57 @@
+// SPDX-License-Identifier: LGPL-3.0-only
+pragma solidity =0.8.19;
+
+import "@openzeppelin/contracts-upgradeable/access/OwnableUpgradeable.sol";
+
+/**
+ * An Azorius extension contract that enables percent based voting basis calculations.
+ *
+ * Intended to be implemented by BaseStrategy implementations, this allows for voting strategies
+ * to dictate any basis strategy for passing a Proposal between >50% (simple majority) to 100%.
```
tbwebb22, line 10: This could work anywhere from 0% to 100% right? Not just 50% to 100%?

contracts/azorius/BaseVotingBasisPercent.sol
```diff
@@ -0,0 +1,57 @@
...
+     * Calculates the basis percentage for a Proposal on the strategy to pass.
+     *
+     * In a simple majority, the basis would be 50% (500,000 / 1,000,000).
+     *
+     * Adjusting the basisNumerator allows for strategies to use any basis
+     * above simple majority, such as 2/3, 90%, etc.
+     */
+    function basis() public view virtual returns (uint256) {
+        return basisNumerator / BASIS_DENOMINATOR;
+    }
```
tbwebb22, line 37: This math doesn't work in Solidity, since all numbers are integers. This would always just return 0, since the numerator is less than the denominator.

contracts/azorius/LinearERC20Voting.sol
tbwebb22: That check is to ensure that the proposal has actually been submitted

contracts/azorius/LinearERC20Voting.sol
tbwebb22, line 179: This math also won't work for the same reason as above. Will probably need to be something like: 

`yesVotes > (yesVotes + noVotes) * basisNumerator / basisDenominator`

contracts/azorius/LinearERC20Voting.sol
tbwebb22: However, our new basis votes check may ensure that proposals that haven't been submitted don't return true for this

tbwebb22 (changes requested):
Looks good, main thing is just the integer math stuff and will need some new tests for this new functionality

contracts/azorius/BaseVotingBasisPercent.sol
tbwebb22: I think we could just remove this function, and use the numerator and denominator in our math directly as I showed in the comment below

contracts/azorius/BaseVotingBasisPercent.sol
herbig: @tbwebb22 updated

contracts/azorius/LinearERC20Voting.sol
```diff
@@ -4,13 +4,14 @@ pragma solidity =0.8.19;
 import { IVotes } from "@openzeppelin/contracts/governance/utils/IVotes.sol";
 import { BaseStrategy, IBaseStrategy } from "./BaseStrategy.sol";
 import { BaseQuorumPercent } from "./BaseQuorumPercent.sol";
+import { BaseVotingBasisPercent } from "./BaseVotingBasisPercent.sol";
```
herbig, line 7: and here

contracts/azorius/BaseVotingBasisPercent.sol
```diff
@@ -0,0 +1,45 @@
+// SPDX-License-Identifier: LGPL-3.0-only
+pragma solidity =0.8.19;
+
+import "@openzeppelin/contracts-upgradeable/access/OwnableUpgradeable.sol";
```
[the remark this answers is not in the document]
tbwebb22: Pushed this change up

contracts/azorius/LinearERC20Voting.sol
tbwebb22: This is resolved, the new `isPassed` logic handles proposals that haven't been submitted yet.

tbwebb22 (approved):

PR #51, opened 2023-03-27
Uint optimize

tbwebb22:
## Description
This PR changes some variables stored as `uint256` to `uint32` for gas savings. Some state variables and structs were re-ordered to take advantage of variable packing.
Some values, such as `quorumNumerator` and `quorumDenominator` did not result in gas savings by changing to a smaller uint, due to the extra variable casting that this causes.

https://github.com/decent-dao/fractal-contracts/issues/41


contracts/azorius/interfaces/IAzorius.sol
```diff
@@ -92,7 +92,14 @@ interface IAzorius {
...
      */
-    function updateTimelockPeriod(uint256 _timelockPeriod) external;
+    function updateTimelockPeriod(uint32 _timelockPeriod) external;
+
+    /**
+     * Updates the execution period for future Proposals.
+     *
+     * @param _executionPeriod new execution period (in blocks)
+     */
+    function updateExecutionPeriod(uint32 _executionPeriod) external;
```
herbig, line 102: oh good call, I'm surprised we didn't have this in here before

herbig (approved):

PR #56, opened 2023-04-04
Change timelocking to check signatures hash rather than transaction hash

herbig:
## Description

Changes timelocking from checking the transaction hash to changing signatures hash.

This prevents skipping timelock by submitting duplicate transaction hashes.


tbwebb22 (approved):

herbig (commented):
looks good!

PR #59, opened 2023-04-23
wrapped voting total supply support

herbig:
## Description

Adds a new voting strategy, an extension of `LinearERC20Voting` which properly uses the total supply of the underlying token when calculating quorum, rather than the total supply of wrapped tokens.


contracts/azorius/LinearERC20Voting.sol
```diff
@@ -151,14 +151,16 @@ contract LinearERC20Voting is BaseStrategy, BaseQuorumPercent, BaseVotingBasisPe
...
+            uint32 endBlock,
+            uint256 votingSupply
         )
     {
         noVotes = proposalVotes[_proposalId].noVotes;
         yesVotes = proposalVotes[_proposalId].yesVotes;
         abstainVotes = proposalVotes[_proposalId].abstainVotes;
         startBlock = proposalVotes[_proposalId].votingStartBlock;
         endBlock = proposalVotes[_proposalId].votingEndBlock;
+        votingSupply = _getProposalVotingSupply(_proposalId);
```
tbwebb22, line 163: `LinearERC20Voting` doesn't actually need this extra storage, right? Adding this is purely for the purpose of supporting the new `LinearERC20WrappedVoting` contract, right? I'm hesitant to add extra storage to this contract for the purpose of making inheritance work for the other one

contracts/LinearERC20WrappedVoting.sol
```diff
@@ -0,0 +1,29 @@
...
+  */
+contract LinearERC20WrappedVoting is LinearERC20Voting {
+
+    /** `proposalId` to "past total supply" of tokens. */
+    mapping(uint256 => uint256) internal votingSupply;
+
+    /** @inheritdoc LinearERC20Voting*/
+    function initializeProposal(bytes memory _data) public virtual override onlyAzorius {
+        super.initializeProposal(_data);
+        votingSupply[abi.decode(_data, (uint32))] = VotesERC20Wrapper(address(governanceToken)).totalSupply();
```
tbwebb22, line 22: I don't think this will work, since the _data variable will include all of the data for the inherited `initializeProposal` data as well. I would think in this `initializePropposal` function you'd want to bring in the logic from the other one, and not call `super.initializeProposal()`

tbwebb22 (commented):
What about this - keeping the struct unmodified, as it currently is. And then on the new `LinearERC20WrappedVoting` contract, add a mapping which is block.number => totalTokenSupply. This gets populated for a given block number when a proposal gets created

contracts/azorius/LinearERC20Voting.sol
herbig: yeah `LinearERC20Voting`, but it's needed to snapshot it when a proposal is created for `LinearERC20WrappedVoting`

All the instances of `LinearERC20Voting` we create on the frontend wouldn't have this storage, only strategies for wrapped tokens would use `LinearERC20WrappedVoting`

herbig: oh hmm wait I did put this on the `LinearERC20Voting`, why did I do that ha

herbig: now I see what you're saying @tbwebb22

contracts/LinearERC20WrappedVoting.sol
tbwebb22: Oh actually disregard my comment here. For some reason I was thinking the decode on `super.initializeProposal` decoded different data than this one, but since they're both decoding to a uint32, this could work.

contracts/azorius/LinearERC20Voting.sol
```diff
@@ -110,7 +110,7 @@ contract LinearERC20Voting is BaseStrategy, BaseQuorumPercent, BaseVotingBasisPe
     }
 
     /** @inheritdoc BaseStrategy*/
-    function initializeProposal(bytes memory _data) external virtual override onlyAzorius {
+    function initializeProposal(bytes memory _data) public virtual override onlyAzorius {
```
herbig, line 113: maybe we should move this down above `hasVoted` now that it's public, the preserve the external / public ordering

tbwebb22 (approved):

contracts/azorius/LinearERC20Voting.sol
```diff
@@ -151,16 +140,29 @@ contract LinearERC20Voting is BaseStrategy, BaseQuorumPercent, BaseVotingBasisPe
...
         noVotes = proposalVotes[_proposalId].noVotes;
         yesVotes = proposalVotes[_proposalId].yesVotes;
         abstainVotes = proposalVotes[_proposalId].abstainVotes;
         startBlock = proposalVotes[_proposalId].votingStartBlock;
         endBlock = proposalVotes[_proposalId].votingEndBlock;
+        votingSupply = getProposalVotingSupply(_proposalId);
     }
 
+    /** @inheritdoc BaseStrategy*/
+    function initializeProposal(bytes memory _data) public virtual override onlyAzorius {
```
herbig, line 156: this function was moved lower in order to sit within the `public` functions, rather than `external`

the only code change was changing it to `public`

PR #60, opened 2023-05-01
Update contracts - add disableInitializers to constructors

tbwebb22:
## Description
This PR:
1) Updates contracts implementing the `FactoryFriendly` pattern, to include the `disableInitializers` function call in constructors, as recommended by the Halborn audit
2) Updates the tests to work with this change
3) Adds a function `quorumVotes` for getting the total number of votes a proposal requires to meet quorum


contracts/azorius/BaseQuorumPercent.sol
```diff
@@ -51,4 +51,12 @@ abstract contract BaseQuorumPercent is OwnableUpgradeable {
...
         return _yesVotes + _abstainVotes >= (_totalSupply * quorumNumerator) / QUORUM_DENOMINATOR;
     }
+
+    /**
+     * Calculates the total number of votes required for a proposal to meet quorum.
+     * 
+     * @param _proposalId The ID of the proposal to get quorum votes for
+     * @return uint256 The quantity of votes required to meet quorum
+     */
+    function quorumVotes(uint32 _proposalId) public view virtual returns (uint256);
```
herbig, line 61: @tbwebb22 should we make this external?

tbwebb22: hmm we could, wondering though if other strategies may have a need to call it internally

herbig: yeah true, good call

herbig (approved):

PR #62, opened 2023-05-26
decent voting strategy initial spec

herbig:
## Description

Initial spec for Decent's voting strategy, which gives voting power for tokens locked in an escrow contract.

### DIFF
--- a/contracts/azorius/LinearERC20Voting.sol
+++ b/contracts/azorius/LinearERC20Voting.sol
@@ -202,7 +202,7 @@ contract LinearERC20Voting is BaseStrategy, BaseQuorumPercent, BaseVotingBasisPe
      * @param _proposalId id of the Proposal
      * @return uint256 the address' voting weight
      */
-    function getVotingWeight(address _voter, uint32 _proposalId) public view returns (uint256) {
+    function getVotingWeight(address _voter, uint32 _proposalId) public view virtual returns (uint256) {
         return
             governanceToken.getPastVotes(
                 _voter,
--- a/contracts/custom/DecentDAOVoting.sol
+++ b/contracts/custom/DecentDAOVoting.sol
@@ -1,0 +1,20 @@
+// SPDX-License-Identifier: LGPL-3.0-only
+pragma solidity =0.8.19;
+
+import { LinearERC20Voting } from "../azorius/LinearERC20Voting.sol";
+import { VotesERC20Wrapper } from "../VotesERC20Wrapper.sol";
+
+interface Escrow {
+    function getBalance(address _address, uint _block) external;
+}
+
+contract DecentDAOVoting is LinearERC20Voting {
+
+    Escrow public escrow;
+
+    // TODO override setUp() function to initialize the escrow contract
+
+    function getVotingWeight(address _voter, uint32 _proposalId) public view override returns (uint256) {
+        return super.getVotingWeight(_voter, _proposalId) + escrow.getBalance(_voter, proposalVotes[_proposalId].votingStartBlock);
+    }
+}


contracts/custom/DecentDAOVoting.sol
```diff
@@ -0,0 +1,20 @@
+// SPDX-License-Identifier: LGPL-3.0-only
+pragma solidity =0.8.19;
+
+import { LinearERC20Voting } from "../azorius/LinearERC20Voting.sol";
+import { VotesERC20Wrapper } from "../VotesERC20Wrapper.sol";
+
+interface Escrow {
+    function getBalance(address _address, uint _block) external;
```
herbig, line 8: @adamgall to implement what we're currently thinking, the escrow contract would just need to have this function.

contracts/custom/DecentDAOVoting.sol
```diff
@@ -0,0 +1,20 @@
...
+}
+
+contract DecentDAOVoting is LinearERC20Voting {
+
+    Escrow public escrow;
+
+    // TODO override setUp() function to initialize the escrow contract
+
+    function getVotingWeight(address _voter, uint32 _proposalId) public view override returns (uint256) {
+        return super.getVotingWeight(_voter, _proposalId) + escrow.getBalance(_voter, proposalVotes[_proposalId].votingStartBlock);
```
herbig, line 18: @tbwebb22 @adamgall does this make sense?  Basically all we should need to do is add the address's true voting weight to the escrow's balance.

This has the issue that it allows users to delegate what they hold themselves, but not sure if that's a problem or not.

contracts/azorius/LinearERC20Voting.sol
```diff
@@ -202,7 +202,7 @@ contract LinearERC20Voting is BaseStrategy, BaseQuorumPercent, BaseVotingBasisPe
      * @param _proposalId id of the Proposal
      * @return uint256 the address' voting weight
      */
-    function getVotingWeight(address _voter, uint32 _proposalId) public view returns (uint256) {
+    function getVotingWeight(address _voter, uint32 _proposalId) public view virtual returns (uint256) {
```
herbig, line 205: since this would require opening this up to virtual, it would be great to iron this out before we deploy the Azorius contracts to mainnet.

contracts/custom/DecentDAOVoting.sol
```diff
@@ -0,0 +1,20 @@
...
+
+interface Escrow {
+    function getBalance(address _address, uint _block) external;
+}
+
+contract DecentDAOVoting is LinearERC20Voting {
+
+    Escrow public escrow;
+
+    // TODO override setUp() function to initialize the escrow contract
```
herbig, line 15: the address of the escrow contract would be supplied in setup()

contracts/azorius/LinearERC20Voting.sol
adamgall: For the record I'm fairly certain that the `virtual` keyword is simply a compile-time sanity check, doesn't affect the compiled bytecode either way.

contracts/custom/DecentDAOVoting.sol
adamgall: > This has the issue that it allows users to delegate what they hold themselves, but not sure if that's a problem or not.

To be clear, is the issue that it allows users to delegate what they hold themselves _but not_ delegate their balance held in the lock contract? So there's like a partial delegation happening?

contracts/azorius/LinearERC20Voting.sol
herbig: oh cool, then all good on timing

contracts/custom/DecentDAOVoting.sol
herbig: yeah, just flagging that that's a possibility

folks would always have a voting weight equal to their locked tokens, but would be able to either self delegate or delegate to someone else any tokens they hold themselves

contracts/azorius/LinearERC20Voting.sol
adamgall: Would be best to double check my assumption here though 😅

contracts/custom/DecentDAOVoting.sol
adamgall: just curious, is it possible to "turn off" delegation for unlocked tokens within this strategy? so that no delegation at all is happening?

herbig: we call `getPastVotes` on the `IVotes` (https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/governance/utils/IVotes.sol) contract, so the ERC20 token would probably need to extended and remove the delegation aspect of the snapshotting.

adamgall: @sethhrbek

Could you please explain more about point 3 of Keeping Delegation? From which point of view are you making that statement? If I own tokens, and delegate the voting power of those tokens to myself or someone else, and then transfer my tokens to another address, the voting power of those tokens does indeed change to whoever the new address has delegated their voting power to.

Also, apologies, but I'm struggling to understand both points in Removing Delegation. You don't need to talk more about it here in this thread, we can catch up next week about it.

contracts/custom/DecentDAOVoting.sol
```diff
@@ -0,0 +1,20 @@
+// SPDX-License-Identifier: LGPL-3.0-only
+pragma solidity =0.8.19;
+
+import { LinearERC20Voting } from "../azorius/LinearERC20Voting.sol";
+import { VotesERC20Wrapper } from "../VotesERC20Wrapper.sol";
+
+interface Escrow {
+    function getBalance(address _address, uint _block) external view returns (uint);
```
adamgall, line 8: The Lock & Release contract allows any token to be locked within it, so could this function take a third parameter for the token address?

```solidity
function getBalance(address _token, address _address, uint _block) external view returns (uint);
```

contracts/custom/DecentVotesERC20.sol
```diff
@@ -0,0 +1,17 @@
...
+pragma solidity =0.8.19;
+
+import { VotesERC20 } from "../VotesERC20.sol";
+
+contract DecentVotesERC20 is VotesERC20 {
+
+    error NotDelegatable();
+
+    function delegates(address account) public view override returns (address) {
+        return account;
```
herbig, line 11: always just returns the account you're requesting as who it's delegated to

contracts/custom/DecentVotesERC20.sol
```diff
@@ -0,0 +1,17 @@
...
+    }
+
+    function _delegate(address, address) internal override {
+        revert NotDelegatable();
```
herbig, line 15: internally if a user attempts to delegate, it just reverts.  This impacts only the public delegate() functions of `ERC20VotesUpgradeable`

https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/master/contracts/token/ERC20/extensions/ERC20VotesUpgradeable.sol

contracts/custom/DecentVotesERC20.sol
herbig: https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/23b9ad8bd15638d2c335bd4ed1c74d825f61dc6a/contracts/token/ERC20/extensions/ERC20VotesUpgradeable.sol#L80

contracts/custom/DecentDAOVoting.sol
herbig: I think the lock contract should return a sum of their voting power, regardless of what tokens they have locked.  The strategy won't know what assets are worth voting power, or what their addresses are

contracts/custom/DecentDAOVoting.sol
herbig: @sethhrbek @adamgall just pushed an example of how to prevent delegation while still maintaining the snapshotting functionality

Also as to point 1, if we did have delegation any vested tokens would be delegated to yourself as soon as they're unlocked, if you have yourself as the delegateee

contracts/custom/DecentDAOVoting.sol
adamgall: Are you saying that if my address holds 10 locked DCNT and 1,000,000 locked SHIT then my voting power is equal to 1,000,010 on Decent's Fractal instance?

That doesn't sound right, or maybe I misinterpreted your suggestion.

The strategy must know the DCNT token address, right? It should be summing the unlocked DCNT token balance in a user's address with the that user's locked DCNT balance in this lock contract.

herbig: ohh, I think I see what you're saying now, I think I misunderstood what you were saying

So escrow will have the `getBalance()` as you suggested, then the strategy will call `getBalance()` with the governance token address

that make sense to me, just pushed up a change to handle that @adamgall

adamgall: Phew, perfect, yes that commit looks like what I was hoping for, thanks!

contracts/custom/DecentDAOVoting.sol
adamgall: do you plan to address this TODO in this PR or another one?

contracts/custom/DecentVotesERC20.sol
adamgall: very clean solution here ser

adamgall:
> One thing to discuss is the mint functionality -- there had been discussion of capping the amount that can be minted in a given time period (like a year).

A while ago Kelvin and I got started on a DCNT Token repo, it includes that "capped mint" functionality (along with a built-in airdrop claim which isn't necessary these days): https://github.com/decent-labs/dcnt

contracts/custom/DecentDAOVoting.sol
herbig: yeah I'll wrap that up next week, I wanted to discuss the setup with @tbwebb22

contracts/azorius/LinearERC20Voting.sol
tbwebb22: I do think you need the `virtual` keyword to be able to override the function on an inheriting contract: 

https://docs.soliditylang.org/en/v0.8.20/contracts.html#function-overriding

contracts/custom/DecentDAOVoting.sol
tbwebb22: Hmm so the `setUp` function in `LinearERC20Voting` would also need to be changed to `virtual`

contracts/custom/DecentVotesERC20.sol
tbwebb22: Nice. We also talked about the voting escrow token being non-transferrable. @adamgall  is that correct?

contracts/custom/DCNTToken.sol
adamgall, line 1: I don't think I want this DCNT token contract in the Fractal repo.

contracts/custom/DCNTToken.sol
```diff
@@ -0,0 +1,137 @@
+//SPDX-License-Identifier: Unlicense
+pragma solidity =0.8.19;
+
+import { IERC20Upgradeable } from "@openzeppelin/contracts-upgradeable/token/ERC20/IERC20Upgradeable.sol";
+import { MerkleProof } from "@openzeppelin/contracts/utils/cryptography/MerkleProof.sol";
+import { UndelegatableVotesERC20 } from "./UndelegatableVotesERC20.sol";
+
+contract DCNTToken is UndelegatableVotesERC20 {
```
adamgall, line 8: But regardless of which repo this contract exists in (https://github.com/decent-dao/fractal-contracts/pull/62/files#r1209304877), I'd like to talk more about this `UndelegatableVotesERC20` inheritance.

I don't want to lock this token into being "undelegatable" forever, just for now.

Earlier when I mentioned that the `delegates` override of just returning `_account` was nice and clean, I didn't realize that this code would be baked into the token contract itself (oversight on my end). 

It's very likely I have a misunderstanding of how everything is hooked together, but I have an assumption that the choice to "acknowledge" delegation lives within a "strategy" that's installed on a Fractal node. So, an ERC20 token could include code that supports delegation (standard `ERC20Votes` functionality), but when the strategy is calculating voting power, it ignores delegation and returns weight based simply on an account's balance at a block. As I type this out, I realize I have an assumption that the standard `ERC20Votes` contract has methods to not only "return an account's delegated-to voting power at some block" _and also_ "return an account's actual balance at some block".

If what I'm saying is unclear please tell me, there's lots of nuance.

herbig: `ERC20Votes` keeps checkpoints of delegated voting power, but not individual balances.

we could definitely use or ignore delegation, and set it up so that it could be a parameter on the strategy, but we would need two checkpoints on the governance token, one for delegation, which `ERC20Votes` already does, and another historical checkpoint added for an individual's token balance

contracts/custom/DecentDAOVoting.sol
```diff
@@ -0,0 +1,111 @@
...
+    function getBalance(address _token, address _holder, uint _block) external view returns (uint);
+}
+
+interface IVotesDecent is IVotes {
+    /**
+     * Returns the token balance that `account` had at a specific moment in the past.
+     * This is distinct from `getPastVotes` in that it is the account's *balance*, not the number of
+     * delegated token votes, which can be different.
+     */
+    function getPastBalance(address account, uint256 timepoint) external view returns (uint256);
```
herbig, line 35: @adamgall this encapsulates whether to use delegation or balance into the strategy, but requires an additional function on the governance token.

contracts/custom/DCNTToken.sol
herbig: @adamgall updated to show how that would work

herbig:
This was useful discussion!

We decided to put token vesting into the Decent token contract itself, and do the voting weight logic there as well.

This should now only require an ERC20 token that implements `IVotes`, and no custom Fractal strategy @adamgall @sethhrbek @tbwebb22 

@adamgall is going to work on that in a separate repo

PR #66, opened 2023-06-06
LinearERC721Voting strategy and ERC721FreezeVoting

herbig:
## Description

Adds a strategy for ERC721 voting, based heavily on the existing Azorius `LinearERC20Voting` strategy, and Usul's [NFT Linear voting](https://github.com/SekerDAO/Usul/blob/master/contracts/votingStrategies/NFTLinearVoting.sol) strategy.

This version however allows multiple NFTs to be registered as governance tokens, each with their own assigned weight.

## Notes

See contract comments for notes on implementation.  Natspec documentation will be more fleshed out when implementation is more settled.

## Testing

TODO, I will write extensive tests after a first round of code review, when implementation is settled.


contracts/azorius/LinearERC721Voting.sol
```diff
@@ -0,0 +1,292 @@
...
+            uint32 startBlock,
+            uint32 endBlock
+        )
+    {
+        noVotes = proposalVotes[_proposalId].noVotes;
+        yesVotes = proposalVotes[_proposalId].yesVotes;
+        abstainVotes = proposalVotes[_proposalId].abstainVotes;
+        startBlock = proposalVotes[_proposalId].votingStartBlock;
+        endBlock = proposalVotes[_proposalId].votingEndBlock;
+        // TODO why didn't we return has voted data in ERC20 strategy?
```
tbwebb22, line 121: hasVoted is a mapping, and a whole mapping can't be returned by a function

contracts/azorius/LinearERC721Voting.sol
```diff
@@ -0,0 +1,292 @@
...
+    function getWeight(
+        uint256 _proposalId,
+        address[] memory _tokenAddresses,
+        uint256[] memory _tokenIds,
+        address _voter
+    ) internal returns (uint256) {
+
+        uint256 weight = 0;
+
+        for (uint256 i = 0; i < _tokenAddresses.length; i++) {
```
tbwebb22, line 151: Could use the 
`unchecked {
 ++i 
}` 
pattern here for some gas savings

contracts/azorius/LinearERC721Voting.sol
```diff
@@ -0,0 +1,292 @@
...
+ */
+contract LinearERC721Voting is BaseStrategy, BaseVotingBasisPercent {
+
+    struct GovernanceNFT {
+        IERC721 token;
+        uint256 weight;
+        bool isProposer;
+    }
+
+    GovernanceNFT[] public governanceTokens;
```
tbwebb22, line 28: I feel like this storage pattern may be more efficient, to prevent having to loop through an array when counting weight:

`
   struct GovernanceNFT {
        uint256 weight;
        bool isProposer;
    }

   mapping(address => GovernanceNFT) public governanceTokens;
`

contracts/azorius/LinearERC721Voting.sol
```diff
@@ -0,0 +1,292 @@
...
+            }
+            
+            for (uint j = 0; j < governanceTokens.length; j++) {
+                GovernanceNFT memory governance = governanceTokens[j];
+                IERC721 token = governance.token;
+                if (address(token) == tokenAddress) {
+                    weight = weight + governance.weight;
+                    break;
+                }
+            }
```
tbwebb22, line 171: This loop could get expensive if there are lots of governance token addresses for voting configured. If the pattern I showed at the top of the contract is used, this loop shouldn't be necessary.

contracts/azorius/LinearERC721Voting.sol
```diff
@@ -0,0 +1,292 @@
...
+
+        if (weight == 0) revert NoVotingWeight();
+
+        _vote(_proposalId, msg.sender, _support, weight);
+    }
+
+    // verifies the voter holds the NFTs and returns the total weight associated with their tokens
+    // the frontend will need to determine whether an address can vote on a proposal, as it is possible
+    // to vote twice if you get more weight later on
+    function getWeight(
```
tbwebb22, line 142: I would think we'd want this function to be available as a `public view` function, and not changing state in any way. To avoid looping through all the tokenAddresses / IDs in two places, maybe this function only takes a single `tokenAddress` and `tokenId`, (and `proposalId & `voter) and then just returns the weight

contracts/azorius/LinearERC721Voting.sol
```diff
@@ -0,0 +1,292 @@
...
+    }
+
+    /** @inheritdoc BaseStrategy*/
+    function isProposer(address _address) public view override returns (bool) {
+        for (uint i = 0; i < governanceTokens.length; i++) {
+            GovernanceNFT memory governance = governanceTokens[i];
+            if (governance.isProposer && governance.token.balanceOf(_address) > 0) {
+                return true;
+            }
+        }
```
tbwebb22, line 250: This loop also could removed if a mapping is used for the governanceTokens data

contracts/azorius/LinearERC721Voting.sol
```diff
@@ -0,0 +1,292 @@
...
+        emit VotingPeriodUpdated(_votingPeriod);
+    }
+
+    /** Internal implementation of `updateQuorumThreshold`. */
+    function _updateQuorumThreshold(uint256 _quorumThreshold) internal {
+        quorumThreshold = _quorumThreshold;
+        emit QuorumThresholdUpdated(quorumThreshold);
+    }
+
+    function _vote(uint32 _proposalId, address _voter, uint8 _voteType, uint256 weight) internal {
```
tbwebb22, line 271: If the `getWeight` function is changed so that it only takes a single token address / ID, the weight lookup could be done inside this function, and weight wouldn't be passed to this function as an argument.

tbwebb22 (commented):
Looks good! Left some comments, main thing I'm thinking about is that `governanceTokens` array, and changing it to a mapping for gas savings.

contracts/azorius/LinearERC721Voting.sol
herbig: oh yeah awesome, we wouldn't really need an array of governance addresses, because voting always involves giving those addresses

I'll have to think through that through but this is probably way better

herbig: hmm, @tbwebb22 we still need a way to check if an individual address can create proposals, in `isProposer(address _address)`
    
    I'm not sure we can do that without keeping an array of NFT addresses and seeing if the personal address has a balance

contracts/azorius/LinearERC721Voting.sol
herbig: @tbwebb22 how can we do this here with just a mapping of governance tokens?  We need to know each token address to check if they have a balance

contracts/azorius/LinearERC721Voting.sol
```diff
@@ -0,0 +1,315 @@
...
+ * of where it was when a proposal was created.
+ *
+ * Also, this uses "quorumThreshold" rather than quorumPercent, because total supply is not knowable
+ * within the IERC721 interface.  This is similar to a multisig "total signers" required, rather than
+ * a percentage of the tokens.
+ */
+contract LinearERC721Voting is BaseStrategy, BaseVotingBasisPercent {
+
+    struct GovernanceNFT {
+        bool exists;
```
tbwebb22, line 23: Could we just use a non-zero weight indicating that it exists?

contracts/azorius/LinearERC721Voting.sol
```diff
@@ -0,0 +1,315 @@
...
+        uint256 weight;
+        bool isProposer;
```
tbwebb22, line 25: Instead of having this as a bool, could we have a `proposerThreshold` and use the weights to determine if a user has enough NFTs to be a proposer?

contracts/azorius/LinearERC721Voting.sol
```diff
@@ -0,0 +1,310 @@
...
+            if (
+                proposalVotes[_proposalId].hasVoted[tokenAddress][tokenId] == true || 
+                _voter != IERC721(tokenAddress).ownerOf(tokenId)
+            ) {
+                continue;
+            }
+            
+            weight += tokenWeights[tokenAddress];
+            proposalVotes[_proposalId].hasVoted[tokenAddress][tokenId] = true;
+        }
```
tbwebb22, line 256: Thoughts on moving this loop into the _vote function? It does seem confusing to me that a function called `_getTotalWeight` isn't a view function, and is changing state

contracts/ERC721FreezeVoting.sol
```diff
@@ -0,0 +1,105 @@
...
+
+    function _getVotesAndUpdateHasVoted(
+        address[] memory _tokenAddresses,
+        uint256[] memory _tokenIds,
+        address _voter
+    ) internal returns (uint256) {
+
+        uint256 votes = 0;
+
+        for (uint256 i = 0; i < _tokenAddresses.length; i++) {
```
tbwebb22, line 87: Could use the unchecked { ++i } pattern here as well

contracts/ERC721FreezeVoting.sol
```diff
@@ -0,0 +1,105 @@
...
+
+        emit ERC721FreezeVotingSetUp(_owner, _strategy);
+    }
+
+    function castFreezeVote() external override pure { revert NotSupported(); }
+
+    function castFreezeVote(address[] memory _tokenAddresses, uint256[] memory _tokenIds) external {
+        if (_tokenAddresses.length != _tokenIds.length) revert UnequalArrays();
+
+        uint256 userVotes = _getVotesAndUpdateHasVoted(_tokenAddresses, _tokenIds, msg.sender);
```
tbwebb22, line 62: So this line is getting called before `freezeProposalCreatedBlock` is getting updated, so it could be counting user votes from the wrong block right?

contracts/ERC721FreezeVoting.sol
herbig: yeah missed that, good call

contracts/ERC721FreezeVoting.sol
herbig: dammit, yeah good catch

I want to have the update be in the same look so I don't need to do it twice, but that might be tricky

test/Azorius-LinearERC721Voting.test.ts
```diff
@@ -0,0 +1,1256 @@
...
+  let holder2Ids: number[];
+  let holder3Ids: number[];
+
+  let mintNFTData: string;
+  let proposalTransaction: TransactionData;
+
+  // Gnosis
+  let createGnosisSetupCalldata: string;
+
+  const gnosisFactoryAddress = "0xa6B71E26C5e0845f74c812102Ca7114b6a896AB2";
```
mudrila: Should we possibly hold it in some const? I believe we wanna use same factory and singleton for all Azorius voting strategy

test/Azorius-LinearERC721Voting.test.ts
```diff
@@ -0,0 +1,1256 @@
...
+      1,
+      ethers.constants.AddressZero,
+      ethers.constants.HashZero,
+      ethers.constants.AddressZero,
+      ethers.constants.AddressZero,
+      0,
+      ethers.constants.AddressZero,
+    ]);
+
+    const predictedGnosisSafeAddress = await predictGnosisSafeAddress(
```
mudrila, line 124: Not sure if it's make sense to refactor right now, but I think we should make same refactoring as we did for FE repo - removing gnosis from terminology

test/Azorius-LinearERC721Voting.test.ts
```diff
@@ -0,0 +1,1256 @@
...
+            60, // timelock period in blocks
+            60, // execution period in blocks
+          ]
+        ),
+      ]);
+
+    await moduleProxyFactory.deployModule(
+      azoriusMastercopy.address,
+      azoriusSetupCalldata,
+      "10031021"
```
mudrila, line 200: Out of readability purposes I think it makes sense to move this nonce to `const`

test/Azorius-LinearERC721Voting.test.ts
```diff
@@ -0,0 +1,1256 @@
...
+              500000, // basis numerator, denominator is 1,000,000, so basis percentage is 50% (simple majority)
+            ]
+          ),
+        ]
+      );
+
+    await moduleProxyFactory.deployModule(
+      linearERC721VotingMastercopy.address,
+      linearERC721VotingSetupCalldata,
+      "10031021" // TODO what's this number???
```
mudrila: That's `nonce`

test/Azorius-LinearERC721Voting.test.ts
```diff
@@ -0,0 +1,1256 @@
...
+
+    it("Basis numerator cannot be updated to a value larger than the denominator", async () => {
+      await expect(
+        linearERC721Voting
+          .connect(gnosisSafeOwner)
+          .updateBasisNumerator(1000001)
+      ).to.be.revertedWith("InvalidBasisNumerator()");
+    });
+
+    it("Only the owner can update the proposer weight on the ERC20LinearVoting", async () => {
```
mudrila: Seems like a typo - not ERC20LinearVoting but ERC721LinearVoting

test/Azorius-LinearERC721Voting.test.ts
```diff
@@ -0,0 +1,1256 @@
...
+
+      // Users vote in support of proposal
+      await expect(
+        linearERC721Voting
+          .connect(tokenHolder1)
+          .vote(0, 1, holder1Tokens, holder1Ids)
+      ).to.be.revertedWith("VotingEnded()");
+    });
+
+    it("A voter cannot vote more than once on a proposal with the same id", async () => {
```
mudrila, line 432: I know we covered this from FE, but might be worth writing test that same user can vote twice but with different IDs

test/MultisigFreezeGuard-ERC721FreezeVoting.test.ts
```diff
@@ -0,0 +1,899 @@
...
+  async function mintNFT(
+    contract: MockERC721,
+    receiver: SignerWithAddress
+  ): Promise<void> {
+    await contract.connect(receiver).mint(receiver.address);
+  }
+
+  beforeEach(async () => {
+    // Fork Goerli to use contracts deployed on Goerli
+    await network.provider.request({
```
mudrila, line 81: This seem to be repeated pattern across all tests - maybe we should move it to some global setup? Or at least make reusable function for this

test/MultisigFreezeGuard-ERC721FreezeVoting.test.ts
```diff
@@ -0,0 +1,899 @@
...
+  buildSafeTransaction,
+  safeSignTypedData,
+  ifaceSafe,
+  abi,
+  predictGnosisSafeAddress,
+  abiSafe,
+  calculateProxyAddress,
+} from "./helpers";
+
+describe("Child Multisig DAO with Azorius Parent", () => {
```
mudrila, line 33: Gonna be tricky, but ideally we should add other DAO types later

test/Azorius-LinearERC721Voting.test.ts
herbig: yeah that's a good idea.. I copied this from other test code, where it's repeated a lot, I'll move it to a constant

test/Azorius-LinearERC721Voting.test.ts
herbig: yeah, I'll leave that for later, but also good call

test/Azorius-LinearERC721Voting.test.ts
herbig: ha, yeah

test/Azorius-LinearERC721Voting.test.ts
herbig: ah, yeah these tests were based on the ERC20 ones so I missed that ha, thanks for catching

test/MultisigFreezeGuard-ERC721FreezeVoting.test.ts
herbig: yeah, all our existing tests repeat almost everything...

I'd love to do a bigger refactor and do what you're describing, but I'll leave that to another PR when I can get to it

PR #83, opened 2024-04-10
Base deployment

### LINKED ISSUE #82: Deploy contracts to Base

adamgall:
closes #82 

- Added Env Vars to support Base
- Added Hardhat config to support Base
- Deployed to Base
- Added Base artifacts to repository


hardhat.config.ts
```diff
@@ -86,14 +88,40 @@ const config: HardhatUserConfig = {
...
+      },
+      {
+        network: "base",
+        chainId: 8453,
+        urls: {
+          apiURL: "https://api.basescan.org/api",
+          browserURL: "https://basescan.org",
+        },
+      },
+    ],
```
adamgall, line 124: Whatever version of `hardhat-verify` we're using doesn't have support for these Base etherscan instances, but luckily we can just add any custom ones we need to. Seemed easier than trying to update dependencies.

adamgall: After doing this, I finished verifying contracts on Base Sepolia (some were already verified...) and verified all contracts on Base that I just deployed.

PR #87, opened 2024-04-14
Update project

### LINKED ISSUE #78: Upgrade Package dependencies
After some discussion, We could benefit from spending some time upgrading the dependencies of this project. 

- Bring `fractal-contracts` repository up to date with current tooling
  - Current `hardhat`
  - `ethers-v6`
- Replace the depreciated `@gnosis.pm` package with the latest from Safe
- Remove everything except `deployments` artifacts from published NPM packages

adamgall:
Closes #78

The following items only affect the repository _around_ the contracts, and are harmless (but great for the health of the repository) to complete.

- [x] Node.js version 20.12.2
- [x] Upgrade project to use current Hardhat 2.22.2.
  - [x] Tests now run in 20 seconds instead of 2 mins yay.
- [x] Update all tests from ethers-v5 to ethers-v6.
- [x] Remove a lot of extra configuration tools and files.
- [x] Trim down published NPM code to just a single `deployments.json` file and the contracts
- [x] Remove a bunch of autogenerated stuff from the repo.

The following items will affect smart contract compilation / logic / behavior. Updates here should be approached cautiously. I actually think we should handle these changes in separate PRs.

- Update `@gnosis.pm/zodiac` package
- Re-include `@gnosis.pm/safe-contracts` package (gives error if in `package.json` at the version it has been at, but doesn't seem needed in there, everything compiles fine)
- Update `@openzeppelin/contracts[-upgradeable]` package[s]

Some important notes:

- Compiling the smart contracts via `npm run compile` (`npx hardhat compile`) still produces the exact same bytecode for all contracts which are currently deployed on all networks.
- The following functionality / features are being _removed_ from the NPM package:
  - TypeChain typing and objects
  - Test files
  - Deployment scripts
- Only Contract addresses and ABIs (per network), and the solidity source code, will remain in the NPM package. However, the format of this data will change.


hardhat.config.ts
```diff
@@ -63,45 +43,53 @@ const config: HardhatUserConfig = {
   networks: {
     mainnet: {
       chainId: 1,
-      url: process.env.MAINNET_PROVIDER || "",
+      url:
+        process.env.MAINNET_PROVIDER || "https://ethereum-rpc.publicnode.com",
       accounts: process.env.MAINNET_DEPLOYER_PRIVATE_KEY
         ? [process.env.MAINNET_DEPLOYER_PRIVATE_KEY]
-        : [],
+        : [dummyPrivateKey],
```
dacebt, line 50: Whats this for?

dacebt: I'm assuming its for testing. Surprised the test environment can't pick up on and deploy with Account 0 in the hardhat node

adamgall: Yeah, kinda goofy.

This is specifically so that the `hardhat verify` task doesn't throw an error. For whatever reason, that task doesn't work unless all of the provider URLs and private keys in this config file are valid values. So, in the common case where you're only putting private keys in your `.env` as quick temporary measures for actually deploying, the task won't run.

dacebt (approved):
LGTM!

PR #93, opened 2024-06-26
DecentHats module

### LINKED ISSUE #92: `[roles-0.1.0]` Create Safe Module for creating full Hats tree

### LINKED ISSUE #94: `[roles-0.1.0]` Create NFT-owned 'smart accounts' for each new Hat
When new Hats are being created, we need to create smart accounts for those Hats along side them.

These smart accounts are ERC6551 accounts, which is a standard for "turning individual NFTs into full-blown identities onchain, with an address, controlled by the specific NFT owner".

Hats is an 1155 contract, so it works for this.

With this feature, we can stream Sablier funds directly to this new smart account address, because for all intents and purposes in this context that address _is_ the Hat.

adamgall:
Closes #92 
Closes #94 

See also https://github.com/decentdao/decent-contracts/pull/91

---

This PR implements a contract called `DecentHats` which will be temporarily added as a module to Safes, in order to create and configure a Hats tree for that Safe.

It also includes support for deploying instances of ERC6551 "Smart Accounts" for each new Hat.

In live environments, we'll use the `HatsAccount1ofN` implementation, which is deployed to various chains at address [0xfEf83A660b7C10a3EdaFdCF62DEee1fD8a875D29](https://github.com/Hats-Protocol/hats-account/releases/tag/0.1.0).

The `ERC6551Registry` also has a canonical address on all chains: [0x000000006551c19487814612e58FE06813775758](https://eips.ethereum.org/EIPS/eip-6551#registry).

### DIFF
--- a/contracts/DecentHats.sol
+++ b/contracts/DecentHats.sol
@@ -19,22 +19,22 @@ contract DecentHats {
     }
 
     IHats public hats;
-    address public keyValuePairs;
-    IERC6551Registry public registry;
     address public hatsAccountImplementation;
+    IERC6551Registry public registry;
+    address public keyValuePairs;
 
     bytes32 public constant SALT = keccak256("DecentHats");
 
     constructor(
         IHats _hats,
-        address _keyValuePairs,
+        address _hatsAccountImplementation,
         IERC6551Registry _registry,
-        address _hatsAccountImplementation
+        address _keyValuePairs
     ) {
         hats = _hats;
-        keyValuePairs = _keyValuePairs;
-        registry = _registry;
         hatsAccountImplementation = _hatsAccountImplementation;
+        registry = _registry;
+        keyValuePairs = _keyValuePairs;
     }
 
     function createAndDeclareTree(
--- a/contracts/DecentHats.sol
+++ b/contracts/DecentHats.sol
@@ -23,7 +23,7 @@ contract DecentHats {
     IERC6551Registry public registry;
     address public hatsAccountImplementation;
 
-    bytes32 public constant SALT = bytes32(abi.encode(0xdece974a75));
+    bytes32 public constant SALT = keccak256("DecentHats");
 
     constructor(
         IHats _hats,
--- a/contracts/DecentHats.sol
+++ b/contracts/DecentHats.sol
@@ -4,6 +4,7 @@ pragma solidity =0.8.19;
 import {Enum} from "@gnosis.pm/zodiac/contracts/core/Module.sol";
 import {IAvatar} from "@gnosis.pm/zodiac/contracts/interfaces/IAvatar.sol";
 import {IHats} from "./interfaces/hats/IHats.sol";
+import {IERC6551Registry} from "./interfaces/IERC6551Registry.sol";
 import {Strings} from "@openzeppelin/contracts/utils/Strings.sol";
 
 contract DecentHats {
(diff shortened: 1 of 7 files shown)


contracts/DecentHats.sol
```diff
@@ -19,10 +20,21 @@ contract DecentHats {
 
     IHats public hats;
     address public keyValuePairs;
+    IERC6551Registry public registry;
+    address public hatsAccountImplementation;
 
-    constructor(IHats _hats, address _keyValuePairs) {
+    bytes32 public constant SALT = bytes32(abi.encode(0xdece974a75));
```
mudrila: Just wondering where this `SALT` is coming from? Maybe that's reasonable to add a comment about it?

contracts/interfaces/IERC6551Registry.sol
```diff
@@ -0,0 +1,51 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.4;
+
+interface IERC6551Registry {
```
mudrila, line 4: Does this mean we have custom registry deployed? Can't we use one from https://github.com/erc6551/reference

contracts/mock/ERC6551Registry.sol
```diff
@@ -0,0 +1,162 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.4;
+
+interface IERC6551Registry {
```
mudrila, line 4: Same question as above - can't we leverage existing package as a dependency?

contracts/mock/MockHatsAccount.sol
```diff
@@ -0,0 +1,22 @@
+// SPDX-License-Identifier: MIT
+pragma solidity =0.8.19;
+
+contract MockHatsAccount {
+    function tokenId() public view returns (uint256) {
+        bytes memory footer = new bytes(0x20);
+        assembly {
+            // copy 0x20 bytes from final word of footer
```
mudrila, line 11: I know that's not related to this PR and rather to Hats Protocol implementation of `HatsAccount` - but why final word of a footer? Just curious

mudrila (commented):
DAMN
THAT's SICK
Straight to prod! 🚀 :shipit:  
Have couple of questions though that are not necessarily needs answering - feel free to resolve those conversations and I'll give my approval

contracts/DecentHats.sol
adamgall: Yeah good call. Just made it up. I'm pretty sure it doesn't really matter, is just used in deterministic address generation via the internal `create2` call that the registry contract makes.

`decenthats` in leetspeek

adamgall: i should probably do something more proper here

adamgall: Updated, idk if this is proper enough though

contracts/interfaces/IERC6551Registry.sol
adamgall: We will not have a custom registry deployed. This Interface is included in our codebase so that I can have proper solidity typing in `DecentHats` when i call the registry to create new Smart Account instances.

contracts/mock/ERC6551Registry.sol
adamgall: This is just a Mock used for Testing

contracts/mock/MockHatsAccount.sol
adamgall: When a new ERC6551 contract is deployed from the Registry, the Registry appends some metadata data to the end of the contract bytecode. That metadata includes the NFT contract address and NFT token ID that _this contract_ belongs to.

So, all we're doing with these functions is exposing those properties.

PR #99, opened 2024-07-24
Update published file

dacebt:
Update published file to just be a javascript file.

### DIFF
--- a/scripts/prepublish.sh
+++ b/scripts/prepublish.sh
@@ -6,7 +6,7 @@ npm run test
 
 npx hardhat export --export-all deployments.json
 
-rm index.ts
+rm index.js
 rm -rf publish && mkdir publish
 
 # Step 1: Remove the `abi`, `name`, and `chainId` keys
@@ -30,7 +30,7 @@ rm abis.json
 
 rm deployments.json 
 
-cat << EOF > index.ts
+cat << EOF > index.js
 import abis from "./publish/abis";
 import addresses from "./publish/addresses";
 export { abis, addresses };
--- a/scripts/prepublish.sh
+++ b/scripts/prepublish.sh
@@ -30,8 +30,8 @@ rm abis.json
 rm deployments.json 
 
 cat << EOF > publish/index.ts
-import abis from "./publish/abis";
-import addresses from "./publish/addresses";
+import abis from "./abis";
+import addresses from "./addresses";
 export { abis, addresses };
 EOF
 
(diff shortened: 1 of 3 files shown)


dacebt:
Whats the publishing workflow? the update package file isn't for bumping to next version. Do you open another PR to do that? or what? @adamgall

adamgall:
@Da-Colon gotta manually create a new Release on GitHub.

Start with creating a tag and pushing it, then create a Release based on that tag.

The new Release triggers the npm publish.

dacebt:
So I bump the version in this PR then?

adamgall:
Oooh sorry I get what you're saying. Since this is in a PR, no. Let's bump the version and tag it directly on develop after this is merged.

If you want to test this out before merging, then ignore what I said above and feel free to bump and tag on this PR.

I'm not going to be too picky about tags it in this repo, it's cool.

dacebt:
@adamgall After looking at it again, I think we need to try the typescript route. because of the `as const` (and not allowed in javascript) allowing for the type to be inferred in the frontend.

scripts/prepublish.sh
```diff
@@ -6,7 +6,6 @@ npm run test
 
 npx hardhat export --export-all deployments.json
 
-rm index.ts
```
adamgall, line 9: this line was here to make sure we're working with a "blank canvas" when this script starts. Maybe instead of removing the line, we change it to `rm publish/index.ts`?

scripts/prepublish.sh
```diff
@@ -30,8 +29,11 @@ rm abis.json
...
-import abis from "./publish/abis";
-import addresses from "./publish/addresses";
+cat << EOF > publish/index.ts
+import abis from "./abis";
+import addresses from "./addresses";
 export { abis, addresses };
 EOF
+
+# Run TypeScript compilation
+tsc ./publish/index.ts
```
adamgall: To be safe (for if this script is run on systems without `tsc` already installed -- seems like Github runners _do_ have it installed already, but that might not be guaranteed in the future), should this line be something like `npx tsc ./publish/index.ts`?

adamgall:
> @adamgall After looking at it again, I think we need to try the typescript route. because of the `as const` (and not allowed in javascript) allowing for the type to be inferred in the frontend.

ooooo yeah good point

scripts/prepublish.sh
dacebt: the line below removes the publish folder and the files within.

scripts/prepublish.sh
dacebt: I can add it to be safe

scripts/prepublish.sh
adamgall: hahaha yes good point!

adamgall (approved):

PR #100, opened 2024-09-05
DecentSablier module

adamgall:
When proposals are open which will affect Sablier streams, there are certain states that blockchain may get into which will render the created proposal invalid at time of execution, despite being valid at time of creation.

All of those scenarios are due to the specific situation in which a stream which has funds to claim from the stream during the proposal creation, no longer has funds to claim from the stream at proposal execution. This can happen because the role member manually flushed out their stream funds during the time when the proposal is active.

In order to protect against these potential race conditions, we should move the logic to determine whether or not to attempt to withdraw funds from a stream from proposal creation time to proposal execution time.

We think a potential solution to this would be to create and deploy a contract with this logic, then:

- ephemerally attach it as a module to the Safe during execution
- call a function on it which includes a list of Sablier streams and maybe some more metadata about those streams
- (if necessary at execution time) proxy a call to flush the Sablier stream through the Safe
- remove the module from the Safe


mudrila (approved):
Looks good as a start, I'm cool with merging it as long as these contracts won't be deployed unless finished.

adamgall:
Oh I don't want to merge this yet, it's nowhere near ready

contracts/interfaces/sablier/LockupLinear.sol
```diff
@@ -25,4 +25,31 @@ library LockupLinear {
...
+        uint40 startTime;
+        uint40 endTime;
+        uint40 cliffTime;
+        bool cancelable;
+        bool wasCanceled;
+        address asset;
+        bool transferable;
+        uint128 totalAmount;
+        address recipient;
+    }
```
adamgall: @DarksightKellar 

Let's make sure that this struct (and any others that we've re-created into our various Sablier Interface files) are fully separated from our _production contracts_. Like, we have our actual contracts, and a handful of Mock contracts to assist with testing. In order to minimize our contract bytecode for the contracts we're deploying to mainnets, let's keep those different structs fully separate and import the necessary files into the necessary contracts.

contracts/mocks/MockSablierV2LockupLinear.sol
```diff
@@ -6,20 +6,7 @@ import "../interfaces/sablier/ISablierV2LockupLinear.sol";
...
-        uint128 totalAmount;
-        address asset;
-        bool cancelable;
-        bool transferable;
-        uint40 startTime;
-        uint40 cliffTime;
-        uint40 endTime;
-    }
-
-    mapping(uint256 => Stream) public streams;
```
adamgall: Ah yeah, maybe it does make sense to keep this struct defined here... (considering my first comment)

contracts/interfaces/sablier/ISablierV2LockupLinear.sol
adamgall, line 1: @DarksightKellar 

So actually, since `DecentHats_0_1_0` uses this interface, and we want to effectively "lock down" that contract from any changes in the future, we're not able to make any changes to this interface: it'll result in any newly compiled versions of `DecentHats_0_1_0` having different bytecode from that which has already been deployed onchain, which will mean our deployment scripts will attempt to re-deploy it and then publish a new contract address in our NPM package. Which will break our front end `smartAccount` address prediction code.

This is definitely frustrating but I'm not sure of any way around it. If you want to make edits to this `ISablierV2LockupLinerar.sol` file, you need to copy the existing code into a new interface file then make the changes there.

You can see what I mean by attempting to run the deployment scripts on Sepolia (with this PR as-is). You'll see that it deploys a new instance of `DecentHats_0_1_0`. We need to avoid that.

DarksightKellar: Yeahh I see what you mean. Alrighty I'll make the necessary changes

adamgall: My recommendation is to create a new interface called `ISablierV2LockupLinearFull.sol`, and just copy in the whole-ass Interface structure from their official repo, to make sure we've got it properly future proofed.

I wasn't really thinking about this when creating the _first_ `ISablierV2LockupLinear.sol` file for `DecentHats_0_1_0`, and so now we've gotta do this. Oops!

edit: might need to the same thing with `LockupLinear.sol`, because there's a reference to that one in `DecentHats_0_1_0`, too.

DarksightKellar: @adamgall  `XYZ_Full.sol` with future-proof code sounds like the best plan (and name lol). Will update

DarksightKellar: @adamgall Yeah this is gonna take a little more thought than a quick copy-paste. Not more difficult, just mismatching variables names and a potential package install (@prb/math/src/UD60x18.sol), or cherry picking which code to copy-paste instead of all.

I'll push what I have meantime, which uses `XYZ.sol` and `XYZ_2.sol` for locked original, and edited code respectively

adamgall: @DarksightKellar i've taken the liberty of finishing this cleanup effort. i think it's all good now

See these commits:
- https://github.com/decentdao/decent-contracts/pull/100/commits/a1e23c469773b30b012fa5427fa063c3cd2c42c4
- https://github.com/decentdao/decent-contracts/pull/100/commits/1ec965046b262037f437088c9ab44690b0c19509
- https://github.com/decentdao/decent-contracts/pull/100/commits/c39571ac6379c136cfb9d24258d6b450d9ceb4f1
- https://github.com/decentdao/decent-contracts/pull/100/commits/a4c4141f287ccc08a35ad90b21318843f25f6240

DarksightKellar (approved):

PR #102, opened 2024-09-17
`New Feature` Termed Roles

### LINKED ISSUE #103: `[Termed Roles]` Create `DecentAutonomousAdminHat` contract

### LINKED ISSUE #104: `[Termed Roles]` Create Tests for `DecentAutonomousAdminHat` contract

### LINKED ISSUE #105: `[Termed Roles]` Update `DecentHat_0_1_0` contract to deploy DecentAAHat as wearer of AdminHat

dacebt:
Adds DecentAutonomousAdmin Contract
- Used for trigging next start and flush any sablier funds to previous elected wearer

Adds DecentHats_0_2_0 Contract
- DecentAutoAdmin clone is deployed and added as wearer of Admin Hat
- Updated to create Termed Roles when prompted
- This involves making the deploying a instance of Hats Election Module and setting the eligibility module

Closes #103 
Closes #104 
Closes #105


contracts/DecentHats_0_1_0.sol
adamgall, line 1: We're going to need to create another version of this contract now (0_2_0), because 

1. this existing version has already been deployed on networks _and_ has been used to create real roles on real daos
2. this contract's address is part of the salt used to determine role's smartAccount addresses.

So this means we can't just update this and deploy a new one, because then we'll be determining the incorrect smartAccount address for any existing Roles on DAOs that used this current version to create their Roles.

adamgall (changes requested):

contracts/DecentHats_0_1_0.sol
dacebt: thats a good point. going to have to figure out a good way to enable versioning of our contracts. Noting for later: What does this transaction to 'update' look like.

contracts/DecentHats_0_1_0.sol
```diff
@@ -247,4 +247,4 @@ contract DecentHats_0_1_0 {
 
         params.hatsProtocol.transferHat(topHatId, address(this), msg.sender);
     }
-}
+}
```
adamgall: Please make sure this change is reverted.

contracts/interfaces/hats/IHats.sol
```diff
@@ -39,4 +39,15 @@ interface IHats {
...
+    function getHatEligibilityModule(
+        uint256 _hatId
+    ) external view returns (address eligibility);
+
+    function isWearerOfHat(
+        address _user,
+        uint256 _hatId
+    ) external view returns (bool isWearer);
+
+    function changeHatEligibility(uint256 _hatId, address _newEligibility) external;
```
adamgall: Please see the following commits and discussions from other PRs:

- Relevant discussion, but for Sablier not Hats: https://github.com/decentdao/decent-contracts/pull/100#discussion_r1792349087
- Relevant commit which implements what I think should happen in this branch, too: https://github.com/decentdao/decent-contracts/pull/108/commits/e1fd7d75a9ba03aa9f47c4fef03a8cf3f359eace

Please let me know if it's not clear why we need to do this. The TL;DR is that if we _don't_, then you'll notice that running the deployment scripts will result in a new instance of `DecentHats_0_1_0` being deployed (because the bytecode has changed), and we need to avoid that.

adamgall (changes requested):

contracts/interfaces/hats/IHats.sol
dacebt: AH you know I watched for that and got fooled by the caching and hadn't looked at the deployment diff yet ha. cool beans I'll take a look at what you've done and update accordingly

adamgall: I actually implemented it on @DarksightKellar's branch... let me cherry pick a couple of commits over

dacebt: even better 🙏

adamgall: @Da-Colon holy shit it ended up being 17 commits LMAO. Had to revert the changes to both the Hats interfaces _and_ the Sablier interfaces. Luckily, the two other open branches implement the "fixes" for both Hats and Sablier, so more of less that part was already done just had to cherry pick them over.

Then, a lot of other commits happened due to cleanup. But that's only because I made changes very atomically. Recommend you read through them in order to see what I did.

contracts/DecentHats_0_2_0.sol
```diff
@@ -11,7 +11,6 @@ import {LockupLinear, Broker} from "./interfaces/sablier/full/types/DataTypes.so
 import {DecentAutonomousAdmin} from "./DecentAutonomousAdmin.sol";
 import {IHatsModuleFactory} from "./interfaces/hats/full/IHatsModuleFactory.sol";
 import {IHatsElectionEligibility} from "./interfaces/hats/full/IHatsElectionEligibility.sol";
-import {Clones} from "@openzeppelin/contracts/proxy/Clones.sol";
```
dacebt: ty

contracts/DecentAutonomousAdmin.sol
```diff
@@ -1,5 +1,6 @@
 // SPDX-License-Identifier: MIT
-pragma solidity =0.8.19;
+pragma solidity 0.8.28;
```
dacebt, line 2: 👍

contracts/DecentHats_0_1_0.sol
```diff
@@ -247,4 +247,4 @@ contract DecentHats_0_1_0 {
 
         params.hatsProtocol.transferHat(topHatId, address(this), msg.sender);
     }
-}
\ No newline at end of file
+}
```
dacebt: 👍

dacebt (commented):
coolio changes makes sense. Question though. Should we just change the 'deploy' script to prevent any new deployments of that contract.

Another question. Should we probably think about a different pattern for updates for V2 and beyond?

contracts/DecentAutonomousAdmin.sol
```diff
@@ -0,0 +1,130 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.28;
+
+import {IHats} from "./interfaces/hats/full/IHats.sol";
+import {IHatsElectionEligibility} from "./interfaces/hats/full/IHatsElectionEligibility.sol";
+import {ISablierV2Lockup} from "./interfaces/sablier/full/ISablierV2Lockup.sol";
+
+contract DecentAutonomousAdmin {
+    string public constant NAME = "DecentAutonomousAdmin";
+    string public version_ = "0.1.0";
```
adamgall: Should probably make this `constant`

adamgall: But actually...

For this contract, which is a "long standing" contract (aka not an ephemeral contract in the context of a proposal execution that we add/remove as a module to execute specific logic), I'd like to see ERC165 support built in. Will be good for having a programmatic way to determine contract functionality when this contract will inevitably be upgraded in the future, and we may have DAOs with different versions of this contract installed on their Admin role.

I think the ERC165 support should replace the (custom, non-idiomatic) `NAME` and `version_` variables.

contracts/DecentHats.sol
```diff
@@ -0,0 +1,324 @@
...
+import {IERC6551Registry} from "./interfaces/IERC6551Registry.sol";
+import {IHats} from "./interfaces/hats/full/IHats.sol";
+import {LockupLinear, Broker} from "./interfaces/sablier/full/types/DataTypes.sol";
+import {DecentAutonomousAdmin} from "./DecentAutonomousAdmin.sol";
+import {IHatsModuleFactory} from "./interfaces/hats/full/IHatsModuleFactory.sol";
+import {IHatsElectionEligibility} from "./interfaces/hats/full/IHatsElectionEligibility.sol";
+import {ModuleProxyFactory} from "@gnosis.pm/zodiac/contracts/factory/ModuleProxyFactory.sol";
+
+contract DecentHats {
+    string public constant NAME = "DecentHats";
```
adamgall: Same comments here. Instead of this NAME variable, I'd like to see ERC165 support built into this contract.

deploy/core/011_deploy_ModuleProxyFactory.ts
adamgall, line 1: I would prefer you undo the changes to this file.

deploy/core/017_deploy_DecentHats_0_1_0.ts
adamgall, line 1: I'd prefer you follow the same pattern as script 011 (after addressing my comment)

test/DecentHats_0_1_0.test.ts
adamgall, line 1: Undo this change please.

adamgall (changes requested):

contracts/DecentHats.sol
adamgall: Wait nevermind we don't need ERC165 in here, this is one of those "ephemeral" contracts.

DarksightKellar (commented):
Man, I started reviewing this some time ago and I remember I had some questions at some point, but now I'm back to it and I don't remember what. Since this is significant work, I'd love to have an all hands engineering review at some point so we can all, both 
 - come to the same page on what the feature+code is and should look like, and
 - contribute and pick apart in real time, where thoughts will bounce of each other and flow a lot more freely.

contracts/DecentAutonomousAdmin.sol
dacebt: Just saw these comments over the weekend. Let me get this done

contracts/DecentAutonomousAdmin.sol
```diff
@@ -0,0 +1,61 @@
...
+        address currentWearer;
+        IHats userHatProtocol;
+        uint256 userHatId;
+        address nominatedWearer;
+    }
+
+    // //////////////////////////////////////////////////////////////
+    //                         initializer
+    // //////////////////////////////////////////////////////////////
+    function setUp(bytes memory initializeParams) public override initializer {}
```
dacebt, line 18: Uses the same patterns as the modules deployed via `ModuleProxyFactory` which requires a setUp function with params, but in its current state we don't need it

contracts/DecentAutonomousAdmin.sol
```diff
@@ -0,0 +1,61 @@
...
+
+    // //////////////////////////////////////////////////////////////
+    //                         Public Functions
+    // //////////////////////////////////////////////////////////////
+    function triggerStartNextTerm(TriggerStartArgs calldata args) public {
```
dacebt, line 23: Pretty straight forward. The Contract allows the Admin Hat, when worn to help trigger the next term. While the next term can be triggered by anyone. we need a few other things to happen when this occurs for UX purposes.

- After starting next term. The Original Hat should be burned via `checkHatWearerStatus`. if ineligibile the hat is burned. Then we just mint a new hat to the new wearer

contracts/DecentHats.sol
```diff
@@ -0,0 +1,395 @@
...
+        (uint256 adminHatId, ) = _createAdminHatAndAccount(
+            params.hatsProtocol,
+            params.registry,
+            params.moduleProxyFactory,
+            params.decentAutonomousAdminMasterCopy,
+            params.hatsAccountImplementation,
+            topHatAccount,
+            topHatId,
+            params.adminHat
+        );
```
dacebt: This function was created from `_createHatAndAccountAndMintAndStreams` and narrowed down to just Admin hat creation.

contracts/DecentHats.sol
```diff
@@ -0,0 +1,393 @@
...
+                    _createElectionEligiblityModule(
+                        params.hatsModuleFactory,
+                        params.hatsElectionEligibilityImplementation,
+                        params.hatsProtocol.getNextId(adminHatId),
+                        topHatId,
+                        params.hats[i].termedParams[0]
+                    ),
+                    adminHatId,
+                    params.hats[i]
+                );
```
dacebt: This is the other 'new' workflow to review

contracts/interfaces/IDecentAutonomousAdmin.sol
```diff
@@ -0,0 +1,21 @@
...
+pragma solidity 0.8.28;
+
+import {IHats} from "./hats/full/IHats.sol";
+import {ISablierV2Lockup} from "./sablier/full/ISablierV2Lockup.sol";
+
+interface IDecentAutonomousAdmin {
+    struct SablierStreamInfo {
+        uint256 streamId;
+        ISablierV2Lockup sablierV2Lockup;
+    }
```
adamgall: I think this struct (and the `ISablierV2Lockup`) are unused and can be removed

contracts/mocks/MockHatsAdmin.sol
adamgall, line 1: Nitpicky but can you rename this file to MockHatsAutoAdmin to match the contract name within

adamgall: Also, what's the reason for this specific mock contract? It seems like it's supposed to mock the core Hats contract. Any reason why you didn't just add this code to the existing MockHats contract?

adamgall (changes requested):
High-level first-draft review

contracts/DecentHats.sol
```diff
@@ -0,0 +1,393 @@
...
+        for (uint256 i = 0; i < params.hats.length; ) {
+            if (params.hats[i].isTermed) {
+                // Create election module and set as eligiblity
+                _createTermedHatAndAccountAndMintAndStreams(
+                    params.hatsProtocol,
+                    topHatAccount,
+                    _createElectionEligiblityModule(
+                        params.hatsModuleFactory,
+                        params.hatsElectionEligibilityImplementation,
+                        params.hatsProtocol.getNextId(adminHatId),
```
adamgall: Has this been tested when there are multiple termed hats being created?

edit: actually I do think this would work as written, yeah. `getNextId(adminHatId)` is being called after each new hat is created, yeah.

test/DecentHats.test.ts
adamgall, line 1: Nowhere in this test are there any hats being created where `isTermed: true`. That seems like an oversight...

adamgall (changes requested):
Please write some tests for the new DecentHats contract where termed hats are being created.

contracts/DecentHats.sol
adamgall: but still, definitely need some tests written

contracts/interfaces/IDecentAutonomousAdmin.sol
dacebt: 🤦 oof yeah good call

test/DecentHats.test.ts
adamgall: @Da-Colon I remember you saying yesterday that there was a reason why there are no tests testing `isTermed: true`. Can you explain that again?

contracts/DecentHats.sol
adamgall, line 1: Make sure that the code from most recent `DecentHats_0_1_0` gets implemented into this contract.

https://github.com/decentdao/decent-contracts/blob/develop/contracts/DecentHats_0_1_0.sol#L198-L233

@DarksightKellar please provide @Da-Colon with the proper natspec comments, too

adamgall (changes requested):

contracts/DecentHats.sol
dacebt: done + updated

contracts/DecentHats.sol
dacebt: got the tree test. Need to add a test for `createTermedRoleHat` external function

contracts/mocks/MockHatsAdmin.sol
dacebt: I'll take another look at the AutoAdmin test and Mock contracts it uses now with everything updated

test/DecentHats.test.ts
```diff
@@ -0,0 +1,826 @@
+import { SignerWithAddress } from '@nomicfoundation/hardhat-ethers/signers';
+import { expect } from 'chai';
+/* eslint-disable-next-line import/no-extraneous-dependencies */
+import { ethers } from 'ethers';
```
dacebt: @adamgall @DarksightKellar So that 'can't find ethers namespace' errors was removing type safety. After quite a bit of trial and error of trying things. I found this is the simplest solution. `hardhat` or one of the `hardhat` libraries is adding `ethers` to our `node_modules` but it is not in our `package.json` but adding it there conflicts with the version used in the hard hat environment. 

So doing this allows for it to properly find it and doesn't prevent the other errors from showing up. 

I did not add this to other files from my own as I didn't want to fill up this PR but I can.

adamgall: I dig it, thanks. I'll open up a standalone PR.

adamgall: Well actually there are only two other test files that have this issue. Would you mind just applying this fix to those two other files in your current PR? thanks.

dacebt: Ah good call thought it was in more places, but looks like its only effecting where we are trying to pull `types` out of the package

test/DecentAutonomousAdmin.test.ts
```diff
@@ -0,0 +1,106 @@
...
+    it('should correctly invalidate random address as current wearer', async function () {
+      const args = {
+        currentWearer: randomUser.address,
+        userHatProtocol: await hatsProtocol.getAddress(),
+        userHatId: userHatId,
+        nominatedWearer: nominatedWearer.address,
+        sablierStreamInfo: [], // No Sablier stream info for this test
+      };
+
+      // Verify the hat is now worn by the current wearer
```
DarksightKellar: Incorrect copy-paste comment

test/DecentAutonomousAdmin.test.ts
```diff
@@ -0,0 +1,106 @@
...
+        userHatId: userHatId,
+        nominatedWearer: nominatedWearer.address,
+        sablierStreamInfo: [], // No Sablier stream info for this test
+      };
+
+      // Call triggerStartNextTerm on the adminHat contract
+      await adminHat.triggerStartNextTerm(args);
+
+      // Verify the hat is now worn by the nominated wearer
+      expect((await hatsProtocol.isWearerOfHat(nominatedWearer.address, userHatId)) === true);
```
DarksightKellar, line 88: Suggest adding another test to verify the previous wearer of the hat was "removed"

DarksightKellar (changes requested):

DarksightKellar (approved):

contracts/DecentHats_0_1_0.sol
adamgall, line 1: Why are there changes in this contract @Da-Colon @DarksightKellar ??

This contract has already been deployed and it's not going to be deployed again (see the commented out deployment code in deployer file 017).

We should not be making changes to contract source code if there's no intention of deploying it. We should make sure that whatever source code is checked into source control always contains the most-recently-deployed contract on public chains.

adamgall (changes requested):

contracts/interfaces/IDecentAutonomousAdmin.sol
```diff
@@ -0,0 +1,16 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.28;
+
+import {IHats} from "./hats/full/IHats.sol";
+import {ISablierV2Lockup} from "./sablier/full/ISablierV2Lockup.sol";
```
adamgall: Does this import need to be here? It's not being referenced in this interface file, so I would say this is probably not the right place for it.

test/DecentHats.test.ts
```diff
@@ -0,0 +1,970 @@
...
+        expect(hatsCountBeforeCreate).to.equal(2); // Top hat + admin hat
+
+        await createRoleHatPromise;
+
+        const newHatId = await mockHats.count();
+        expect(newHatId).to.equal(3); // + newly created hat
+      });
+    });
+
+    describe('Creating a new hat on existing Tree', () => {
```
DarksightKellar: Should be "Creating a new Termed hat on existing tree"

contracts/mocks/MockHats.sol
adamgall, line 1: Considering that this Mock contract implements the not-"`full`" `IHats` interface, which is specifically _only_ used in an old version of `DecentHats_0_1_0`, I'm wondering why there needs to be additional functionality added to this Mock contract?

contracts/DecentHats_0_1_0.sol
DarksightKellar: Reverted. The natspec update went onto the wrong contract

contracts/mocks/MockHats.sol
adamgall, line 1: Why do we need additional functionality in this contract? This specific Mock implements the non-"`full`" `IHats` interface, which is only used in the old `DecentHats_0_1_0` contract. I do see that there are failing tests if I remove these new mocked functions, but that leads me to believe that the test file is relying on this `MockHats` instance when maybe it should be relying on the new `MockHatsAutoAdmin` instance.

contracts/mocks/MockHatsAdmin.sol
adamgall, line 1: Please make sure the filename of this contract matches the actual contract name.

adamgall, line 1: Is there any reason why this Mock file / contract has the word "Admin" in it?

Maybe a more appropriate name would be something like "MockHatsFull", to refer to the fact that it implements the "full" IHats interface.

adamgall (changes requested):

contracts/DecentHats_0_1_0.sol
adamgall: Thank you

contracts/mocks/MockHatsAdmin.sol
adamgall: fixed

adamgall: fixed

contracts/mocks/MockHats.sol
adamgall: fixed

contracts/interfaces/IDecentAutonomousAdmin.sol
adamgall: fixed

test/DecentHats.test.ts
adamgall: fixed

contracts/mocks/MockHatsAdmin.sol
adamgall: fixed

contracts/DecentHats.sol
adamgall: addressed

contracts/DecentAutonomousAdmin.sol
```diff
@@ -0,0 +1,51 @@
...
+        require(
+            args.hatsProtocol.isWearerOfHat(args.currentWearer, args.hatId),
+            "Not current wearer"
+        );
```
adamgall: To stick with modern solidity, can we use actual `error`s here?

contracts/DecentAutonomousAdmin.sol
```diff
@@ -0,0 +1,51 @@
...
+        address hatsEligibilityModuleAddress = args
+            .hatsProtocol
+            .getHatEligibilityModule(args.hatId);
```
adamgall: To save some gas, let's not allocate memory for this variable, since it's only being used in one place. Just inline this into line 33.

contracts/DecentHats.sol
```diff
@@ -0,0 +1,485 @@
...
+pragma solidity 0.8.28;
+
+import {Enum} from "@gnosis.pm/safe-contracts/contracts/common/Enum.sol";
+import {IAvatar} from "@gnosis.pm/zodiac/contracts/interfaces/IAvatar.sol";
+import {Strings} from "@openzeppelin/contracts/utils/Strings.sol";
+import {IERC20} from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import {IERC6551Registry} from "./interfaces/IERC6551Registry.sol";
+import {IHats} from "./interfaces/hats/full/IHats.sol";
+import {LockupLinear, Broker} from "./interfaces/sablier/full/types/DataTypes.sol";
+import {DecentAutonomousAdmin} from "./DecentAutonomousAdmin.sol";
```
adamgall: This import seems unneeded.

contracts/DecentHats.sol
adamgall: We also probably just don't even need this NAME string in the contract either. Like who cares, yanno? Idk why I added it in DecentHats_0_1_0.

contracts/DecentHats.sol
```diff
@@ -0,0 +1,485 @@
...
+
+    struct Hat {
+        address wearer;
+        string details;
+        string imageURI;
+        SablierStreamParams[] sablierParams;
+        TermedParams[] termedParams;
+        uint32 maxSupply;
+        bool isMutable;
+        bool isTermed;
```
adamgall: Do we need to have this boolean? Can we not infer it based on the existence of any items in the `termedParams` array?

contracts/DecentHats.sol
```diff
@@ -0,0 +1,485 @@
...
+        uint128 termEndDateTs;
+        address[] nominatedWearers;
+    }
+
+    struct Hat {
+        address wearer;
+        string details;
+        string imageURI;
+        SablierStreamParams[] sablierParams;
+        TermedParams[] termedParams;
```
adamgall: Why is this an array?

contracts/DecentHats.sol
```diff
@@ -0,0 +1,485 @@
...
+        IHatsModuleFactory hatsModuleFactory,
+        address hatsElectionEligibilityImplementation,
+        uint256 hatId,
+        uint256 topHatId,
+        TermedParams calldata termedParams
+    ) internal returns (address electionModuleAddress) {
+        electionModuleAddress = hatsModuleFactory.createHatsModule(
+            hatsElectionEligibilityImplementation,
+            hatId,
+            abi.encode(topHatId, uint256(0)),
```
adamgall: Does this 0 need to be cast? Also, can you please add some comments on these lines about what these "immutable" and "mutable" args represent, from the PoV of the Elections Eligibility contract?

contracts/DecentHats.sol
```diff
@@ -0,0 +1,485 @@
...
+        Hat calldata hat
+    ) internal {
+        require(
+            hat.termedParams.length == 1,
+            "DecentHats: termedParams length must be 1"
+        );
+        require(
+            hat.termedParams[0].nominatedWearers.length == 1,
+            "DecentHats: nominatedWearers length must be 1"
+        );
```
adamgall: Can we use actual `error`s here? And everywhere in this contract.

contracts/DecentHats.sol
```diff
@@ -0,0 +1,485 @@
...
+            topHatAccount,
+            eligibilityAddress
+        );
+
+        IHatsElectionEligibility(eligibilityAddress).elect(
+            hat.termedParams[0].termEndDateTs,
+            hat.termedParams[0].nominatedWearers
+        );
+
+        hatsProtocol.mintHat(hatId, hat.termedParams[0].nominatedWearers[0]);
```
adamgall: Definitely open to pushback on this suggestion, but, what if we called `batchMintHat` here and passed in the entire `nominatedWearers` array? Leave it to the frontend to just make sure we're only passing in one of them. That might allow us to continue using this same contract if we do ever decide to open up "role members" to > 1 at a time.

contracts/DecentHats.sol
```diff
@@ -0,0 +1,485 @@
...
+            adminHatId,
+            hat,
+            topHatAccount,
+            eligibilityAddress
+        );
+
+        IHatsElectionEligibility(eligibilityAddress).elect(
+            hat.termedParams[0].termEndDateTs,
+            hat.termedParams[0].nominatedWearers
+        );
```
adamgall: Nitpick, flip the order of these two statements.

contracts/DecentHats.sol
adamgall: Edit: nevermind, this would cause a discrepancy in how we're setting up streams then, cause we are hardcoding the [0] there. And it's beyond us to decide how streams should happen if there are multiple nominated wearers.

So now maybe I'm thinking, can we just turn `nominatedWearers` into a single `address`? We'd need to put that address into an array when calling eligibility.elect, but that's fine. I kind of like that better, making the contract's input params match our business logic.

contracts/DecentHats.sol
```diff
@@ -0,0 +1,481 @@
...
+            _hatsProtocol.createHat(
+                adminHatId,
+                _hat.details,
+                _hat.maxSupply,
+                eligibility,
+                toggle,
+                _hat.isMutable,
+                _hat.imageURI
+            );
+    }
```
adamgall: Is it benefitting us to have this single external function call wrapped in an internal function? Why not just inline the `_hatsProtocol.crateHat()` call wherever it's used?

contracts/DecentHats.sol
```diff
@@ -0,0 +1,481 @@
...
+    ) internal returns (address) {
+        return
+            _registry.createAccount(
+                _hatsAccountImplementation,
+                SALT,
+                block.chainid,
+                protocolAddress,
+                hatId
+            );
+    }
```
adamgall: Same as previous comment

contracts/DecentHats.sol
```diff
@@ -0,0 +1,481 @@
...
+        TermedParam calldata termedParam
+    ) internal returns (address electionModuleAddress) {
+        electionModuleAddress = hatsModuleFactory.createHatsModule(
+            hatsElectionEligibilityImplementation,
+            hatId,
+            abi.encode(topHatId, uint256(0)),
+            abi.encode(termedParam.termEndDateTs),
+            uint256(SALT)
+        );
+    }
```
adamgall: same as previous comment

adamgall (changes requested):

contracts/DecentHats.sol
adamgall: fixed, thanks

contracts/DecentHats.sol
adamgall: fixed

contracts/DecentHats.sol
adamgall: fixed

contracts/DecentHats.sol
adamgall: addressed

contracts/DecentHats.sol
adamgall: addressed this

contracts/DecentAutonomousAdmin.sol
dacebt: fixed

contracts/DecentAutonomousAdmin.sol
dacebt: fixed

contracts/DecentHats.sol
dacebt: ?

dacebt: I guess it doesn't really matter, but it seem to make sense to create the hat then elect?

contracts/DecentHats.sol
dacebt: fixed.

contracts/DecentHats.sol
dacebt: fixed.

contracts/DecentHats.sol
dacebt: fixed.

contracts/DecentHats.sol
adamgall: yeahh... my nitpick is just because the `hatId` variable isn't used in the `elect` call, but it's used in the line below that. My intention with the nitpick comment was just to keep relevant code close to each other.

> I guess it doesn't really matter, but it seem to make sense to create the hat then elect?

but the `elect`ion doesn't depend on the new `hatId`, is what i'm saying.

contracts/DecentHats.sol
```diff
@@ -0,0 +1,436 @@
...
+            params.adminHatId,
+            params.hat
+        );
+
+        params.hatsProtocol.transferHat(
+            params.topHatId,
+            address(this),
+            msg.sender
+        );
+    }
```
adamgall: I think the last thing to do in this contract, is clean up this standalone function.

@DarksightKellar this one is on you.

So, the reason why we need to transfer the top back back to the safe, is because _this_ contract is the current wearer of the top hat by the time this function is being called. We did that (transaction was created in the frontend and part of the bundle of txs in the proposal that this call is in) because this contract can't create a new hat on a tree unless it's an admin of the hat it's trying to create. Makes sense.

BUT, instead of transferring the top hat to this contract, why not utilize the fact that we've added this contract as a module on the Safe! As of right now, adding this contract as a module on the Safe (in the context of minting a new hat via this call), _is completely unnecessary_, _unless we actually call `execTransactionFromModule` in this function call_.

So let's do that instead of transferring the top hat around.

Please proxy the call to create a new hat through the Safe via `execTransactionFromModule`, and remove the `transferHat` instruction. In the frontend, no need to transfer the top hat to this contract.

contracts/DecentHats.sol
```diff
@@ -0,0 +1,436 @@
...
```
adamgall: Oh look this function exists too!

@Da-Colon @DarksightKellar see my previous comment in this PR. Same applies here.

contracts/DecentHats.sol
adamgall: fixed

adamgall (changes requested):
Last set of change requests, related to the two external standalone functions which create new Hats on existing trees.

Those functions _shouldn't_ need to transfer the top hat back to the Safe, because this contract _shouldn't_ need to take ownership of the top hat in the first place.

Just use the fact that we add this contract as a module on the Safe, to proxy those calls to create new hats through the Safe.

contracts/DecentHats.sol
dacebt: Let's talk about this today.

adamgall: ready when you are 🫡

adamgall (approved):

PR #108, opened 2024-10-02
Whitelisted proposers

### LINKED ISSUE #107: Create LinearERC20VotingWithHatsProposalCreation strategy
See https://github.com/orgs/decentdao/projects/32/views/3?pane=issue&itemId=81926660

### LINKED ISSUE #110: CreateLinearERC721VotingWithHatsProposalCreation strategy
See https://github.com/orgs/decentdao/projects/32/views/3?pane=issue&itemId=81926660

adamgall:
Closes #107 
Closes #110

Implements two new Strategy contracts
- Whitelisted Hats able to create proposals with ERC20 token voting
- Whitelisted Hats able to create proposals with ERC721 token voting


contracts/azorius/LinearERC20VotingExtensible.sol
```diff
@@ -0,0 +1,327 @@
+// SPDX-License-Identifier: LGPL-3.0-only
+pragma solidity =0.8.19;
+
+import {IVotes} from "@openzeppelin/contracts/governance/utils/IVotes.sol";
+import {BaseStrategy, IBaseStrategy} from "./BaseStrategy.sol";
+import {BaseQuorumPercent} from "./BaseQuorumPercent.sol";
+import {BaseVotingBasisPercent} from "./BaseVotingBasisPercent.sol";
+
+/**
+ * An [Azorius](./Azorius.md) [BaseStrategy](./BaseStrategy.md) implementation that
```
mudrila: Could you please update these comments to outline specific details about this `Extensible` strategy?

contracts/azorius/LinearERC20VotingWithHatsProposalCreation.sol
```diff
@@ -0,0 +1,152 @@
...
+            }
+        }
+        if (!found) revert HatNotWhitelisted();
+
+        emit HatRemovedFromWhitelist(_hatId);
+    }
+
+    /** @inheritdoc LinearERC20VotingExtensible*/
+    function isProposer(address _address) public view override returns (bool) {
+        for (uint256 i = 0; i < whitelistedHatIds.length; i++) {
```
mudrila: So eventually you're leaning towards OR but not EITHER case - right? In that case - do we wanna that "emergency strategy for swapping strategies"?

mudrila: I'm still convinced that EITHER case is more feasible option

contracts/azorius/LinearERC20VotingWithHatsProposalCreation.sol
```diff
@@ -0,0 +1,152 @@
...
+                0, // requiredProposerWeight is zero because we only care about the hat check
+                _quorumNumerator,
+                _basisNumerator
+            )
+        );
+
+        if (_hatsContract == address(0)) revert InvalidHatsContract();
+        hatsContract = IHats(_hatsContract);
+
+        for (uint256 i = 0; i < _initialWhitelistedHats.length; i++) {
```
mudrila: Shall we require that `_initialWhitelistedHats` has at least 1 hat? Otherwise strategy will be locked

mudrila (changes requested):
Have couple requests and questions.
Also, seems like we're lacking deployment script for new strategy

contracts/azorius/LinearERC20VotingWithHatsProposalCreation.sol
adamgall: yes we should! Currently having trouble getting this associated test written though...

contracts/azorius/LinearERC20VotingExtensible.sol
adamgall: sure!

contracts/azorius/LinearERC20VotingWithHatsProposalCreation.sol
adamgall: This is fixed now

mudrila (approved):

contracts/azorius/HatsProposalCreationWhitelist.sol
```diff
@@ -0,0 +1,119 @@
...
+
+    /**
+     * @dev Checks if an address is authorized to create proposals.
+     * @param _address The address to check for proposal creation authorization.
+     * @return bool Returns true if the address is wearing any of the whitelisted Hats, false otherwise.
+     * @notice This function overrides the isProposer function from the parent contract.
+     * It iterates through all whitelisted Hat IDs and checks if the given address
+     * is wearing any of them using the Hats Protocol.
+     */
+    function isProposer(address _address) public view virtual returns (bool) {
```
DarksightKellar: The purpose of `isProposer` wasn't clear to me when reading the tests until I came to read the doc on this function. Whatcha think about renaming to something like `canPropose` or `isWearerOfWhitelistedHat`?

DarksightKellar (approved):
Lookin' good. Had one observation

contracts/azorius/HatsProposalCreationWhitelist.sol
mudrila: It is called `isProposer` to keep consistency with other voting strategies and `BaseStrategy` - so yeah, have to keep this naming 🤷🏼‍♂️

contracts/azorius/LinearERC20VotingWithHatsProposalCreation.sol
```diff
@@ -0,0 +1,82 @@
+// SPDX-License-Identifier: LGPL-3.0-only
+pragma solidity =0.8.19;
+
+import {LinearERC20VotingExtensible} from "./LinearERC20VotingExtensible.sol";
+import {HatsProposalCreationWhitelist} from "./HatsProposalCreationWhitelist.sol";
+import {IHats} from "../interfaces/hats/IHats.sol";
+
+/**
+ * An [Azorius](./Azorius.md) [BaseStrategy](./BaseStrategy.md) implementation that
+ * enables linear (i.e. 1 to 1) ERC21 based token voting, with proposal creation
```
dacebt: Typo. `ERC21`

contracts/azorius/HatsProposalCreationWhitelist.sol
dacebt, line 1: Instead of `whitelistedHatIds`, could we instead use a mapping like `mapping(uint256 => bool) public whitelistedHats;` 

This would allow for us to doconstant-time (O(1)) checks. And it either exists or is true/false.

dacebt: This would require an additional state like `whiteListCount` to update as well. but make alot of this cleaned up as well

contracts/azorius/LinearERC20VotingWithHatsProposalCreation.sol
dacebt, line 1: What are the purpose of these proposal creation contracts?

adamgall: These are the new strategy contracts that support whitelisted proposal creation with standard token voting.

contracts/azorius/HatsProposalCreationWhitelist.sol
adamgall: @mudrila what does the client need? Will we ever need to show a "list" of all whitelisted hats? should we be keeping _both_ an array _and_ a mapping?

adamgall: Answering my own question here... I don't see any usages of `getWhitelistedHatsCount` or `whitelistedHatIds` in the frontend codebase (on the `feature/whitelisting` branch), which leads me to assume that we do _not_ need to worry about "list out all whitelisted Hats", and so we probably can just swap out this "array" structure with a "mapping" structure and probably also _don't_ need to keep a separate count of whitelisted hats.

cc @Da-Colon @mudrila sounds good?

contracts/azorius/LinearERC20VotingWithHatsProposalCreation.sol
adamgall: Fixed

contracts/azorius/HatsProposalCreationWhitelist.sol
adamgall: Ah, wait, hmmmmm. The `isProposer` function. Because it accepts an `address`, (not a `uint hatId`, we would still need to loop through all of the whitelisted hat IDs and check if the given address is a wearer of each of those hats.

is it acceptable to add another parameter (uint256 hatId) to the `isProposer` function? Does this function need to follow a certain interface? idk, will fuck around and find out.

adamgall: Yeah, `isProposer` needs to have exactly the signature `isProposer(address)`, because that's what Azorius is expecting is implemented on each strategy contract. So... ok, back to the drawing board.

In order for `isProposer(address)` to work, the contract will _need_ to know a list of all whitelisted Hat IDs. No way around that. And if it needs to know the full set, a `mapping` won't cut it. This is why it was implemented as an `array` in the first place. @Da-Colon

PR #127, opened 2024-11-15
don't mint if wearer isn't changing

dacebt:


adamgall (changes requested):
Please update comments (on line 30, too), and i would like to see some test changes/updates/additions in this PR too please.

adamgall (approved):

PR #137, opened 2025-02-06
ENG-174 Added an IVersion protocol and its implementations on the voting strategies.

johnqh:
Q: What's our smart contract change process? I don't expect to deploy to main net until merge to main, but how about Sepolia?


contracts/azorius/interfaces/IVersion.sol
```diff
@@ -0,0 +1,13 @@
...
+
+/**
+ * Interface of functions for versioned contracts
+ */
+interface IVersion {
+    /**
+     * Returns the current version of the contract
+     *
+     */
+    function getVersion() external pure returns (uint8);
```
adamgall: Maybe we actually do want this type to be larger? I guess there could be a world in which we exceed 256 versions.

adamgall:
A: Usually what we do is that the engineer working on the new contracts will simply deploy to sepolia (`npx hardhat deploy --network sepolia`). Don't commit the changes (artifacts) yet. Hardcode the new contract addresses into the frontend for testing, until we're ready for a full deployment on all networks, after which we publish a new version of the `@fractal-framework/fractal-contracts` package and bump it in the interface.

contracts/azorius/interfaces/IVersion.sol
johnqh: Changed to uint16 in new code

johnqh:
Instead of modifying existing file, I created new classes with V2 postfix. This is important because our FE needs to know the address of all master contract versions.

adamgall (changes requested):
We'll probably want some additional deployment scripts for these new contracts as well.

johnqh:
> We'll probably want some additional deployment scripts for these new contracts as well.

Added 021_ to 024_ ... under "deploy" folder. Is there anything else I need to do?

adamgall:
> > We'll probably want some additional deployment scripts for these new contracts as well.
> 
> Added 021_ to 024_ ... under "deploy" folder. Is there anything else I need to do?

Nope, that's all! When it comes time to actually do a full deployment of these contracts on all of the networks, then we'd add the deployment artifacts to the repo.

adamgall (changes requested):
I don't think we need the "Extensible" flavors of these new V2 contracts.

I had to create those at the time that we made the "Hats" versions of the strategy contracts, because the original strategy contracts didn't have any of their functions marked as `virtual`, so I wasn't able to override them in the child "Hats" contracts (specifically the `setUp` and `isProposer` functions). My workaround was to simply copy/paste the original strategy contracts into new contracts and then add the `virtual` keyword to all functions.

The fact that the original strategy contracts did not have the `virtual` keyword was a mistake. So since we're creating these new V2 flavors, I think we should remove the "Extensible" name _and_ mark the functions in those new base V2 contracts as `virtual`.

When looking at this new PR, my gut's telling me that we can probably avoid creating the "ExtensibleV2" flavors of contracts.

adamgall (approved):

contracts/interfaces/IVersion.sol
```diff
@@ -0,0 +1,13 @@
+//SPDX-License-Identifier: MIT
+pragma solidity ^0.8.19;
```
adamgall, line 2: Set the version to `^0.8.19` instead of `=0.8.19`, so that any inheriting contracts can use any solidity version up to 0.9.0.

dacebt (approved):

DarksightKellar (approved):
Sweet

PR #138, opened 2025-02-14
ENG-172 ERC4337 Get owner

johnqh:
Changed voting strategy to get owner from the msg.sender if it is the smart account.

```
  ERC4337VoterSupport
    voter
      when the msgSender is a smart account
        ✔ should return the owner of the smart account
        ✔ should return address(0) when smart account owner is zero address
      when the msgSender is an EOA
        ✔ should return the msgSender
      when the msgSender is a contract that does not implement IOwnership
        ✔ should return the contract address
```

### DIFF
--- a/contracts/azorius/strategies/v2/LinearERC20VotingV2.sol
+++ b/contracts/azorius/strategies/v2/LinearERC20VotingV2.sol
@@ -3,16 +3,28 @@ pragma solidity =0.8.19;
 
 import {LinearERC20VotingExtensible} from "../LinearERC20VotingExtensible.sol";
 import {IVersion} from "../../../interfaces/IVersion.sol";
-
+import {ERC4337VoterSupport} from "./ERC4337VoterSupport.sol";
 /**
  * An [Azorius](./Azorius.md) [BaseStrategy](./BaseStrategy.md) implementation that
  * enables linear (i.e. 1 to 1) token voting. Each token delegated to a given address
  * in an `ERC20Votes` token equals 1 vote for a Proposal.
  */
-contract LinearERC20VotingV2 is LinearERC20VotingExtensible, IVersion {
+
+contract LinearERC20VotingV2 is LinearERC20VotingExtensible, IVersion, ERC4337VoterSupport {
     /** @inheritdoc IVersion*/
     function getVersion() external pure virtual returns (uint16) {
         // This should be incremented whenever the contract is modified
         return 2;
+    }
+
+    /** @inheritdoc LinearERC20VotingExtensible*/
+    function vote(uint32 _proposalId, uint8 _voteType) external virtual override {
+        address voter = _voter(msg.sender);
+        _vote(
+            _proposalId,
+            voter,
+            _voteType,
+            getVotingWeight(voter, _proposalId)
+        );
     }
 }
--- a/contracts/azorius/strategies/v2/LinearERC20VotingV2.sol
+++ b/contracts/azorius/strategies/v2/LinearERC20VotingV2.sol
@@ -4,13 +4,17 @@ pragma solidity =0.8.19;
 import {LinearERC20VotingExtensible} from "../LinearERC20VotingExtensible.sol";
 import {IVersion} from "../../../interfaces/IVersion.sol";
 import {ERC4337VoterSupport} from "./ERC4337VoterSupport.sol";
+
 /**
  * An [Azorius](./Azorius.md) [BaseStrategy](./BaseStrategy.md) implementation that
  * enables linear (i.e. 1 to 1) token voting. Each token delegated to a given address
  * in an `ERC20Votes` token equals 1 vote for a Proposal.
  */
-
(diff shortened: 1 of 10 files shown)


contracts/azorius/strategies/v2/LinearERC20VotingV2.sol
```diff
@@ -3,16 +3,28 @@ pragma solidity =0.8.19;
...
         // This should be incremented whenever the contract is modified
         return 2;
     }
+
+    /** @inheritdoc LinearERC20VotingExtensible*/
+    function vote(uint32 _proposalId, uint8 _voteType) external virtual override {
+        address voter = _voter(msg.sender);
+        _vote(
+            _proposalId,
+            voter,
```
DarksightKellar, line 32: Might as well inline these calls to `_voter(msg.sender)`, no? Slight gas savings, I think

adamgall (commented):
Do we want to support the situation where both the Smart Account contract _and_ the owner's EOA own tokens?

If this implementation goes live, and a user attempts to vote with their Smart Account, then any voting power that has been delegated to the Smart Account will be lost.

Also, please run `npm run pretty` on these changes. I see (at least one) syntax issues.

contracts/azorius/strategies/v2/LinearERC20VotingV2.sol
adamgall: Doubt it would be much, it's all part of the same contract.

johnqh:
> Do we want to support the situation where both the Smart Account contract _and_ the owner's EOA own tokens?
> 
> If this implementation goes live, and a user attempts to vote with their Smart Account, then any voting power that has been delegated to the Smart Account will be lost.
> 

Do it in a different ticket? This would require to get the voting weight from the smart contract first, and only try to get from parent if smart contract voting weight is 0. It makes sense but I need to have the IVotingWeight interface (ENG-158) first, and this will create bigger code changes. I also want to unblock UX development.

Created ENG-248 for this.


> Also, please run `npm run pretty` on these changes. I see (at least one) syntax issues.

Done that but I didn't see any changes?

contracts/azorius/strategies/v2/LinearERC20VotingV2.sol
johnqh: Inlined in the ERC721 contracts. The ERC20 has two places using voter (the second one calling getVotingWeight), so I think the variable should save gas.

adamgall:
> Done that but I didn't see any changes?

You know what, it's not `prettier`.

I have the "Solidity" extension by "NomicFoundation" installed (the team behind Hardhat), and apparently there's some kind of "format on save" thing built into that extension for solidity files.

These are the changes that occur on my machine when I open the new file you created and then just save it.

```diff
diff --git a/contracts/azorius/strategies/v2/ERC4337VoterSupport.sol b/contracts/azorius/strategies/v2/ERC4337VoterSupport.sol
index eb6e90e..2fba89d 100644
--- a/contracts/azorius/strategies/v2/ERC4337VoterSupport.sol
+++ b/contracts/azorius/strategies/v2/ERC4337VoterSupport.sol
@@ -16,13 +16,13 @@ abstract contract ERC4337VoterSupport {
     /**
      * Returns the address of the voter which owns the voting weight
      * @param _msgSender address of the sender. It can be the wallet address, or the smart account address with EOA as owner
-     * 
+     *
      */
     function _voter(address _msgSender) internal virtual returns (address) {
-        try IOwnership(_msgSender).owner()  returns (address _value) {
+        try IOwnership(_msgSender).owner() returns (address _value) {
             return (_value);
-        } catch  {
+        } catch {
             return _msgSender;
         }
     }
-}
\ No newline at end of file
+}
```

Would you mind installing this same extension?

I wonder if there's a way to get this formatting job into a package.json script.

johnqh (commented):
Inlined _voter(...) call in LinearERC721VotingV2 and LinearERC721VotingWithHatsProposalCreationV2

johnqh:
@adamgall Strange. I have this extension installed and also Formatter set to "prettier". It doesn't do the auto-formatting for me. I am going to dig a little more after coffee. Meanwhile, do you mind formatting it for me?

adamgall:
> Meanwhile, do you mind formatting it for me?

No problem. Pushed.

adamgall:
> Do it in a different ticket? This would require to get the voting weight from the smart contract first, and only try to get from parent if smart contract voting weight is 0. It makes sense but I need to have the IVotingWeight interface (ENG-158) first, and this will create bigger code changes. I also want to unblock UX development.

Ok yeah, we can do that (edge case considerations) in a separate ticket, if the issue here is simply to unblock UX.

adamgall (changes requested):
Would be good to see some new tests in this PR considering we have functional contract changes.

adamgall (approved):

dacebt (approved):
Looks straightforward

PR #139, opened 2025-02-17
Deploy simple account factory

DarksightKellar:
Closes [ENG-183](https://linear.app/decent-labs/issue/ENG-183/create-the-smart-account-contract-with-the-ability-to-return-the-owner)

All added files are from code on [this 0.7.0 release branch](https://github.com/eth-infinitism/account-abstraction/blob/releases/v0.7/contracts/samples/SimpleAccountFactory.sol).

Because we couldn't find a version of `SimpleAccountFactory` deployed that points to `EntryPoint` 0.7.0, I've gone ahead and deployed this `SimpleAccountFactory` to Sepolia with the 0.7.0 `EntryPoint` address.


adamgall:
Received a response from Dror Tirosh, regarding if there's already a deployment of SimpleAccountFactory:

> SimpleAccount and its factory are samples, not official accounts to use.
> It comes to demonstrate how to use 4337, not to be a full fledged account. 
> Eg it misses some features an account should have, such as erc1271 support.
>
> That said, you can freely deploy a factory and use it.
> There is no "master" factory.

And @DarksightKellar @johnqh honestly i think he has a good point regarding ERC1271. It seems like a generally good idea to build support for that into these SimpleAccount contracts.

The use case being that in the future we will likely need users to "sign in" to the Decent App using their wallet by creating signatures, and if we're implementing these smart accounts now maybe we should prepare for that future in which the Smart Account becomes the first class wallet for a given user in our app.

DarksightKellar:
> Received a response from Dror Tirosh, regarding if there's already a deployment of SimpleAccountFactory:
> 
> > SimpleAccount and its factory are samples, not official accounts to use.
> > It comes to demonstrate how to use 4337, not to be a full fledged account.
> > Eg it misses some features an account should have, such as erc1271 support.
> > That said, you can freely deploy a factory and use it.
> > There is no "master" factory.
> 
> And @DarksightKellar @johnqh honestly i think he has a good point regarding ERC1271. It seems like a generally good idea to build support for that into these SimpleAccount contracts.
> 
> The use case being that in the future we will likely need users to "sign in" to the Decent App using their wallet by creating signatures, and if we're implementing these smart accounts now maybe we should prepare for that future in which the Smart Account becomes the first class wallet for a given user in our app.

@adamgall We could always deploy a new implementation contract and have smart account proxies upgrade to it, for when such an upgrade is necessary, right?

Mayyybe to reduce friction, the `_authorizeUpgrade` can be changed to allow for the DAO/dao-governance to upgrade.

EDIT:
Had a chat with @adamgall. We're gonna leave this PR unmerged just so FE would have something on Sepolia to test (SimpleAccount has been deployed. see deployments folder for address).

After this contracts repo has been updated, and before an official deploy, we'll want to create an extended version of `SimpleAccount` that implements ERC1271.

DarksightKellar:
Closing because no longer needed

PR #140, opened 2025-02-18
Add ERC165 to v2 strategies

DarksightKellar:
Closes [ENG-173](https://linear.app/decent-labs/issue/ENG-173/add-erc-165-support-on-strategy-contracts)

Initial thoughts, feel like we need to add some more interfaces? Will dig into this but pushing this meantime for visibility


adamgall (approved):
I would agree that we should create interfaces for the specific functions that exist in these contracts, then make sure those interfaces are included in the `supportsInterface` functions.

But we can add those in the future. Specifically, as part of the work I'm currently doing, I'm going to re-look at the ERC165 support for _all_ contracts, and I'll get all of this standardized and consistent.

dacebt (approved):

DarksightKellar:
Closing because this will be included as part of Adam's ongoing contracts upgrade+refactoring work

PR #142, opened 2025-02-18
Feature/eng 158 votingweight interface

johnqh:
This is an attempt to move the source of truth of voting weight to the strategies.

However, I am thinking about moving the code to a functional contract (such as VotingWeightHelper) instead of interface implementations. Not sure what the best pattern is so open for discussion.

### DIFF
--- a/contracts/azorius/strategies/v2/ERC721VotingWeightSupport.sol
+++ b/contracts/azorius/strategies/v2/ERC721VotingWeightSupport.sol
@@ -1,0 +1,88 @@
+// SPDX-License-Identifier: LGPL-3.0-only
+pragma solidity =0.8.19;
+
+import {ERC721VotingToken, ERC721VotingWeight, IERC721VotingWeight} from "../../../interfaces/IERC721VotingWeight.sol";
+import {LinearERC721VotingExtensible} from "../LinearERC721VotingExtensible.sol";
+import {IERC721} from "@openzeppelin/contracts/token/ERC721/IERC721.sol";
+
+abstract contract ERC721VotingWeightSupport {
+    function _votingWeight(
+        address _address,
+        mapping(address => uint256) storage _tokenWeights,
+        address[] memory _tokenAddresses,
+        uint256[] memory _tokenIds
+    ) internal view returns (uint256) {
+        uint256 weight;
+
+        for (uint256 i = 0; i < _tokenAddresses.length; ) {
+            address tokenAddress = _tokenAddresses[i];
+            uint256 tokenId = _tokenIds[i];
+
+            require(
+                _address == IERC721(tokenAddress).ownerOf(tokenId),
+                "Not owned"
+            );
+            weight += _tokenWeights[tokenAddress];
+            unchecked {
+                ++i;
+            }
+        }
+        return weight;
+    }
+
+    function _unusedVotingPower(
+        address _address,
+        uint32 _proposalId,
+        mapping(address => uint256) storage _tokenWeights,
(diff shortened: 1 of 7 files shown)


johnqh:
Second thought, this should be implemented as an interface within the strategy for ENG-248.

DarksightKellar:
Should this be based on develop? I'm seeing code from the other PR here

adamgall:
@DarksightKellar fixed

contracts/azorius/strategies/v2/LinearERC20VotingV2.sol
```diff
@@ -34,4 +36,24 @@ contract LinearERC20VotingV2 is
...
+
+    /** @inheritdoc IERC20VotingWeight*/
+    function votingWeight(
+        address _address
+    ) external view override returns (uint256) {
+        return governanceToken.getVotes(_address);
+    }
+
+    /** @inheritdoc IERC20VotingWeight*/
+    function unusedVotingWeight(
```
DarksightKellar, line 48: "unused"?

johnqh:
> Should this be based on develop? I'm seeing code from the other PR here

Updated

contracts/azorius/strategies/v2/LinearERC20VotingV2.sol
johnqh: Suggestions for better names? How about availableVotingWeight?

adamgall: I like "available", yeah

DarksightKellar: ~Yep sounds good~  Wait no this is actually a deferred call to `governanceToken.getPastVotes` which is getting voting power _at the time the propsal is created_ (`proposalVotes[_proposalId].votingStartBlock`). That fact should reflect somehow.

How about something like `getVotingWeightForProposal`? `getProposalVotingWeight`?

I'd also suggest updating to `votingWeight` to `getCurrentVotingWeight`.

EDIT: Or honestly just `getVotingWeight` should be fine to replace `unusedVotingWeight` lol since its `_proposalId` param kinda implies ties to a specific proposal. Then `getCurrentVotingWeight` differentiates and identifies itself enough. Thoughts?

contracts/azorius/strategies/v2/ERC721VotingWeightSupport.sol
```diff
@@ -0,0 +1,88 @@
...
+        }
+
+        ERC721VotingWeight memory result;
+        result.weight = weight;
+        if (count == _tokenAddresses.length) {
+            result.tokens = tokens;
+        } else {
+            // Condense the result array
+            ERC721VotingToken[] memory unvoted = new ERC721VotingToken[](count);
+            for (uint32 i = 0; i < count; i++) {
```
DarksightKellar, line 80: Wondering about the use of `uint32` VS `uint256` up on line `48`

contracts/azorius/strategies/v2/ERC721VotingWeightSupport.sol
```diff
@@ -0,0 +1,88 @@
...
+            );
+            weight += _tokenWeights[tokenAddress];
+            unchecked {
+                ++i;
+            }
+        }
+        return weight;
+    }
+
+    function _unusedVotingPower(
```
DarksightKellar, line 33: Looks like you forgot to update this to `_unusedVotingWeight`

PR #144, opened 2025-02-20
New Directory Structure

adamgall:
Autoformat and begin to move contracts around into new directory structure.

There are zero functional changes in here, just moving contracts around and making sure they're formatted correctly.

All tests still pass with zero modifications to them.

This isn't the _final_ directory structure, but it begins to but some better patterns into place.

This PR targets `develop`, and "changes the bytecode" of pretty much every contract. So, if we merge this into `develop`, we'll be effectively blocked from deploying any new contracts from this repo until the whole bulk of upcoming refactoring work is finished.

However... if we did run into a situation where we need to urgently get a new contract deployed, we could always just manually edit the deployment scripts to skip re/deploying existing contracts that had changes.

So yeah, let's just keep this targeted at `develop` and run with it.

PS I'm not going to attach a linear issue to this, because there are more "directory structure cleanup" changes coming later. I had just unfortunately done this file-moving-around as one of the first things in my big chain of commits, so I'm pretty much unable to rebase these commits around.


johnqh (approved):
I am OK with "This PR targets develop, and "changes the bytecode" of pretty much every contract. So, if we merge this into develop, we'll be effectively blocked from deploying any new contracts from this repo until the whole bulk of upcoming refactoring work is finished."

"However... if we did run into a situation where we need to urgently get a new contract deployed, we could always just manually edit the deployment scripts to skip re/deploying existing contracts that had changes."

We can also create a release branch from "main" for hot fixes, same as in decent-interface

adamgall:
> We can also create a release branch from "main" for hot fixes, same as in decent-interface

We don't have a main branch in this repo, develop is the trunk lololol

PR #155, opened 2025-02-20
License cleanup

adamgall:
- Modify all contracts to use `// SPDX-License-Identifier: AGPL-3.0`
  - AGPL-3.0 was originally used in some of the Hats interfaces (directly copy/pasted into this project), and is pretty restrictive, so if we want to utilize this code in our codebase we need to at least match it.
  - Technically we can make _only_ the contracts which inherit from Hats be this AGPL-3.0 license, and then use a less restrictive one for other contracts, but in the spirit of consistency why not just make em all the same yanno.
- Doesn't touch some of the Sablier interfaces (which were straight copy/pasted into this project), which have a `// SPDX-License-Identifier: GPL-3.0-or-later` identifier, and is compatible with AGPL-3.0

### DIFF
--- a/contracts/deployables/account-abstraction/BasePaymasterV1.sol
+++ b/contracts/deployables/account-abstraction/BasePaymasterV1.sol
@@ -1,4 +1,4 @@
-// SPDX-License-Identifier: GPL-3.0
+// SPDX-License-Identifier: AGPL-3.0
 pragma solidity ^0.8.28;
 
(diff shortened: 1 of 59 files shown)


DarksightKellar (approved):
Wondered about the details, asked the GPT for a quick summary:

Screenshot 2025-02-20 at 23 34 07

LGTM

johnqh (approved):

PR #184, opened 2025-03-09
Add ERC165 support across multiple contracts

adamgall:
Depends on https://github.com/decentdao/decent-contracts/pull/183

Implements comprehensive ERC165 interface support for various contracts:
- Add `supportsInterface` method to multiple contracts
- Dynamically calculate and support interface IDs
- Enhance contract interoperability and introspection capabilities
- Update test suites to validate interface support
- Add utility function for calculating interface IDs

The changes improve contract flexibility and provide a standardized way to query supported interfaces across the project's contracts.

PR #185, opened 2025-03-10
Introduce Version abstract contract and standardize versioning across contracts

adamgall:
Depends on https://github.com/decentdao/decent-contracts/pull/184

Implements a new Version abstract contract to provide a consistent versioning mechanism:
- Create Version contract with ERC165 interface support
- Replace IVersion interface with a more flexible implementation
- Update multiple contracts to inherit from Version
- Modify getVersion method to be view instead of pure
- Remove redundant IVersion interface checks in supportsInterface methods
- Add comprehensive test suite for Version contract

The changes improve contract versioning, enhance interface support, and provide a more flexible approach to contract identification and introspection.

### DIFF
--- a/contracts/deployables/account-abstraction/DecentPaymasterV1.sol
+++ b/contracts/deployables/account-abstraction/DecentPaymasterV1.sol
@@ -3,16 +3,13 @@ pragma solidity ^0.8.28;
 
 import {BasePaymasterV1, IEntryPoint} from "./BasePaymasterV1.sol";
 import {IDecentPaymasterV1} from "../../interfaces/decent/deployables/IDecentPaymasterV1.sol";
-import {IVersion} from "../../interfaces/decent/deployables/IVersion.sol";
+import {Version} from "../Version.sol";
 import {PackedUserOperation, IPaymaster} from "@account-abstraction/contracts/interfaces/IPaymaster.sol";
 import {ERC165} from "@openzeppelin/contracts/utils/introspection/ERC165.sol";
 
-contract DecentPaymasterV1 is
-    IDecentPaymasterV1,
-    IVersion,
-    BasePaymasterV1,
-    ERC165
-{
+contract DecentPaymasterV1 is IDecentPaymasterV1, Version, BasePaymasterV1 {
+    uint16 private constant VERSION = 1;
+
     // Mapping: strategy address => function selector => is approved
     mapping(address => mapping(bytes4 => bool)) public approvedFunctions;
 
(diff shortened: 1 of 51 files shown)


contracts/deployables/account-abstraction/DecentPaymasterV1.sol
```diff
@@ -3,16 +3,16 @@ pragma solidity ^0.8.28;
...
 
-contract DecentPaymasterV1 is
-    IDecentPaymasterV1,
-    IVersion,
-    BasePaymasterV1,
-    ERC165
-{
+contract DecentPaymasterV1 is IDecentPaymasterV1, Version, BasePaymasterV1 {
+    // Version identification constants
+    string private constant CONTRACT_FAMILY = "account-abstraction";
```
johnqh: I like the additional functions. Is there a way to have a string enum with Solidity (or const string)? 

We don't want the binary to be difference when we add new enum or const string to the file. If that's not possible, We can have a VersionHelperV1 utility class for now, with the known contract families and implementation types. When we have new values, we use a new VersionHelperV2 instead of modifying VersionHelperV1

adamgall: Enums in solidity just map to integers.

May I ask what issues you think we'll encounter by leaving these properties as naked strings inlined to the contracts like this?

johnqh: I see multiple CONTRACT_FAMILY with the same value, such as "freeze-guard". Since each class defines CONTRACT_FAMILY as a string, there is room for typos. This becomes more important when we encourage third party developers to develop their own strategies. They will look at CONTRACT_FAMILY and IMPLEMENTATION_TYPE and wonder what the significances are, so we need to explain their meaning and how they are used by the FE. This is easier if we have a single file with all the possible values.

adamgall: Hmm

> I see multiple CONTRACT_FAMILY with the same value, such as "freeze-guard".

Yes

> Since each class defines CONTRACT_FAMILY as a string, there is room for typos.

True, but I don't think it really *matters* if there's a typo, tbh. All that matters is that a client's developers are able to look at the solidity code, understand how the `contractId` is calculated, and calculate it.

Alternatively we could probably just pre-compute the `contractIds` and hardcode them in the contract.

> This becomes more important when we encourage third party developers to develop their own strategies.

I mean technically, anyone could implement a contract which returns any value for `getContractId`, even a value that some other contract has used already. I don't think there's much we can do about that in the current contract architecture/patterns.

If a third party developer makes their own strategy, they could also just use random gibberish for those three properties (family/type/version), and it's fine, as long as the client is aware of those values.

> They will look at CONTRACT_FAMILY and IMPLEMENTATION_TYPE and wonder what the significances are, so we need to explain their meaning and how they are used by the FE. This is easier if we have a single file with all the possible values.

I do agree with this. Maybe this is solved with better inline comments/documentation?

...

I'm also thinking about implementing some kind of singleton factory contract that can create (upgradable) instances of all of our contracts. Maybe there's some kind of "registration" that happens in that contract first, so developers can claim specific `contractIds`, effectively tying them to a specific implementation?

Idk, I think more thought needs to go into this. I'm not entirely satisfied with this current implementation.

johnqh (approved):

PR #188, opened 2025-03-13
Voting weight interface

johnqh:
So FE can retrieve voting weight from contract, instead of complicated calculations


twodam (approved):
lgtm

adamgall (changes requested):
IMO we should have a new test file for the `ERC721VotingWeightSupportV1` contract, and probably some modifications to the `...VotingV1` test files too

contracts/deployables/strategies/ProposalVotesV1.sol
dacebt, line 1: Is this valid? I don't believe you can define a struct with a mapping without storage (contained within library or contract).

adamgall:
Closing this because this work isn't relevant any longer, with rebuild of Strategy and Voting/Proposer adapters.

PR #203, opened 2025-03-24
Refactor BaseStrategyV1 and related contracts to replace Azorius module with Proposal Initializer

adamgall:
- Updated BaseStrategyV1 to use a proposalInitializer instead of azoriusModule for authorization.
- Removed the setAzorius function and associated event, replacing it with a new initialize function that accepts a proposalInitializer.
- Modified LinearERC20VotingV1 and LinearERC721VotingV1 to align with the new proposalInitializer logic.
- Updated tests to reflect changes in initialization and authorization checks.

These changes enhance the clarity and functionality of the strategy contracts, ensuring proper access control for proposal initialization.

### DIFF
--- a/contracts/deployables/strategies/LinearERC20VotingV1.sol
+++ b/contracts/deployables/strategies/LinearERC20VotingV1.sol
@@ -96,7 +96,7 @@ contract LinearERC20VotingV1 is
      *
      * @param _owner Address that will own the contract
      * @param _governanceToken The token used for voting
-     * @param _azoriusModule Address of the Azorius module contract
+     * @param _proposalInitializer Address that is allowed to initialize Proposals
      * @param _votingPeriod Time period for voting
      * @param _requiredProposerWeight Minimum weight to create proposals
      * @param _quorumNumerator Numerator for quorum calculation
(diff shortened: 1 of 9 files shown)


contracts/deployables/strategies/LinearERC20VotingV1.sol
```diff
@@ -117,7 +117,7 @@ contract LinearERC20VotingV1 is
         governanceToken = IVotes(_governanceToken);
         __Ownable_init(_owner);
         __UUPSUpgradeable_init();
-        _setAzorius(_azoriusModule);
```
dacebt, line 120: Something doesn't feels off by this. requiring each strategy to have this line of code for something that is required for the contract to function properly. 

I guess my question comes down to, is there value in those "non 0" 
checks prior to setting the azorius module? if yes, then we should probably keep the internal method and reuse that? 

if no, should we lock all of our strategies to this pattern? Would we ever want something else to authorize starting a proposal?

dacebt (approved):

contracts/deployables/strategies/LinearERC20VotingV1.sol
adamgall: Addressed in latest (force pushed) commit

twodam (approved):

PR #211, opened 2025-03-28
Add HatsProposalCreationWhitelistV1 contract and associated interfaces

adamgall:
depends on https://github.com/decentdao/decent-contracts/pull/210

- Introduced `HatsProposalCreationWhitelistV1` contract for managing proposal creation permissions based on whitelisted Hats.
- Created `IHatsProposalCreationWhitelistV1` interface defining the contract's functionality and events.
- Developed `ConcreteHatsProposalCreationWhitelistV1` mock contract for testing purposes.
- Added `MockHats2` contract to simulate the Hats protocol for testing.
- Implemented comprehensive tests for `HatsProposalCreationWhitelistV1`, covering initialization, whitelisting, unwhitelisting, and proposal authorization logic.

### DIFF
--- a/contracts/azorius/strategies/HatsProposalCreationWhitelistV1.sol
+++ b/contracts/azorius/strategies/HatsProposalCreationWhitelistV1.sol
@@ -1,0 +1,115 @@
+// SPDX-License-Identifier: AGPL-3.0
+pragma solidity ^0.8.19;
+
+import {IHatsProposalCreationWhitelistV1} from "../../interfaces/IHatsProposalCreationWhitelistV1.sol";
+import {IHats} from "../../interfaces/hats/IHats.sol";
+import {OwnableUpgradeable} from "@openzeppelin/contracts-upgradeable/access/OwnableUpgradeable.sol";
+import {ERC165} from "@openzeppelin/contracts/utils/introspection/ERC165.sol";
+
+abstract contract HatsProposalCreationWhitelistV1 is
+    IHatsProposalCreationWhitelistV1,
+    OwnableUpgradeable,
+    ERC165
+{
+    IHats public hatsContract;
+
+    /** Array to store whitelisted Hat IDs. */
+    uint256[] private whitelistedHatIds;
+
+    constructor() {
+        _disableInitializers();
+    }
+
+    /**
+     * Sets up the contract with its initial parameters.
+     *
+     * @param initializeParams encoded initialization parameters:
+     * `address _hatsContract`, `uint256[] _initialWhitelistedHats`
+     */
+    function setUp(bytes memory initializeParams) public virtual {
+        (address _hatsContract, uint256[] memory _initialWhitelistedHats) = abi
+            .decode(initializeParams, (address, uint256[]));
+
+        if (_hatsContract == address(0)) revert MissingHatsContract();
+        hatsContract = IHats(_hatsContract);
+
+        if (_initialWhitelistedHats.length == 0) revert NoHatsWhitelisted();
+        for (uint256 i = 0; i < _initialWhitelistedHats.length; i++) {
+            _whitelistHat(_initialWhitelistedHats[i]);
+        }
+    }
+
+    /**
+     * Adds a Hat to the whitelist for proposal creation.
+     * @param _hatId The ID of the Hat to whitelist
+     */
+    function whitelistHat(uint256 _hatId) external onlyOwner {
+        _whitelistHat(_hatId);
+    }
+
+    /**
+     * Internal function to add a Hat to the whitelist.
+     * @param _hatId The ID of the Hat to whitelist
+     */
+    function _whitelistHat(uint256 _hatId) internal {
+        for (uint256 i = 0; i < whitelistedHatIds.length; i++) {
+            if (whitelistedHatIds[i] == _hatId) revert HatAlreadyWhitelisted();
+        }
+        whitelistedHatIds.push(_hatId);
+        emit HatWhitelisted(_hatId);
+    }
+
+    /**
+     * Removes a Hat from the whitelist for proposal creation.
+     * @param _hatId The ID of the Hat to remove from the whitelist
+     */
+    function unwhitelistHat(uint256 _hatId) external onlyOwner {
+        bool found = false;
+        for (uint256 i = 0; i < whitelistedHatIds.length; i++) {
+            if (whitelistedHatIds[i] == _hatId) {
+                whitelistedHatIds[i] = whitelistedHatIds[
+                    whitelistedHatIds.length - 1
+                ];
+                whitelistedHatIds.pop();
+                found = true;
+                break;
+            }
+        }
+        if (!found) revert HatNotWhitelisted();
+
+        emit HatUnwhitelisted(_hatId);
+    }
+
+    /**
+     * @dev Checks if an address is authorized to create proposals.
+     * @param _address The address to check for proposal creation authorization.
+     * @return bool Returns true if the address is wearing any of the whitelisted Hats, false otherwise.
+     * @notice This function overrides the isProposer function from the parent contract.
+     * It iterates through all whitelisted Hat IDs and checks if the given address
+     * is wearing any of them using the Hats Protocol.
+     */
+    function isProposer(address _address) public view virtual returns (bool) {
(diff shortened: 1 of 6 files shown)


contracts/interfaces/IHatsProposalCreationWhitelistV1.sol
```diff
@@ -0,0 +1,72 @@
...
+     * @dev Returns the Hats contract.
+     * @return The Hats contract interface.
+     */
+    function hatsContract() external view returns (IHats);
+
+    /**
+     * @dev Adds a Hat to the whitelist for proposal creation.
+     * @param _hatId The ID of the Hat to whitelist
+     */
+    function whitelistHat(uint256 _hatId) external;
```
twodam, line 52: would `addHatToWhitelist` be more unify to style of `removeHatFromWhitelist`? or maybe use `whitelistHat` with `unwhitelistHat`?

contracts/azorius/strategies/HatsProposalCreationWhitelistV1.sol
```diff
@@ -0,0 +1,115 @@
...
+     *
+     * @param initializeParams encoded initialization parameters:
+     * `address _hatsContract`, `uint256[] _initialWhitelistedHats`
+     */
+    function setUp(bytes memory initializeParams) public virtual {
+        (address _hatsContract, uint256[] memory _initialWhitelistedHats) = abi
+            .decode(initializeParams, (address, uint256[]));
+
+        if (_hatsContract == address(0)) revert InvalidHatsContract();
+        hatsContract = IHats(_hatsContract);
```
twodam: it can still be **invalid** hats contract even if it pass the check of zero address.

but I guess we has no reliable method to validate this? `!ERC165(_hatsContract).supportsInterface(type(IHats).interfaceId)` won't work because Hats contract itself didn't add that interfaceId to `supportsInterface`?

how about add something like this?
```solidity
(bool success, ) = _hatsContract.staticcall(
     abi.encodeWithSignature("isWearerOfHat(address,uint256)", address(this), 0)
);
require(success, "InvalidHatsContract");
```

contracts/azorius/strategies/HatsProposalCreationWhitelistV1.sol
```diff
@@ -0,0 +1,115 @@
...
+     */
+    function getWhitelistedHatIds() public view returns (uint256[] memory) {
+        return whitelistedHatIds;
+    }
+
+    function supportsInterface(
+        bytes4 interfaceId
+    ) public view virtual override returns (bool) {
+        return
+            interfaceId == type(IHatsProposalCreationWhitelistV1).interfaceId ||
```
twodam, line 112: nice! I don't know why Hats didn't add interfaceId of their own `IHats` interface.

test/HatsProposalCreationWhitelistV1.test.ts
```diff
@@ -0,0 +1,309 @@
...
+      ).to.be.revertedWithCustomError(getModuleProxyFactory(), 'FailedInitialization');
+    });
+
+    it('should not allow initialization with invalid hats contract', async () => {
+      // Deploy another implementation for this test
+      const mockImplementation = await new ConcreteHatsProposalCreationWhitelistV1__factory(
+        deployer,
+      ).deploy();
+
+      // Create initialization data with zero address for hats contract
```
twodam, line 153: add test case where maybe someone accidentally pass a `ERC20` contract (or a contract guarantee to have no `isWearerOfHat` function) as hats contract? what's the consequence if someone did this? update to new strategy with correct hats contract?

test/HatsProposalCreationWhitelistV1.test.ts
```diff
@@ -0,0 +1,309 @@
...
+        .connect(owner)
+        .removeHatFromWhitelist(proposerHatId1);
+
+      const whitelistedHats = await concreteHatsProposalCreationWhitelist.getWhitelistedHatIds();
+      expect(whitelistedHats.length).to.equal(1);
+      expect(whitelistedHats[0]).to.equal(proposerHatId2);
+    });
+  });
+
+  describe('ERC165', () => {
```
twodam, line 280: As per the ERC-165 spec, no interface should ever match `0xffffffff`.

image

twodam (changes requested):

contracts/interfaces/IHatsProposalCreationWhitelistV1.sol
adamgall: yes I agree, will make this change

contracts/azorius/strategies/HatsProposalCreationWhitelistV1.sol
adamgall: although everything you're saying is true and accurate...

the zero address check is more about making sure that the parameter hasn't been accidentally omitted during the `setUp` call. considering that this `setUp` call is only made once per instance, and our own app is the one creating that calldata, I'm not overly concerned about this. i will probably not change this code.

contracts/azorius/strategies/HatsProposalCreationWhitelistV1.sol
adamgall: my assumption for why they didn't implement ERC165 in their contract: is because the Hats contract is a singleton instance onchain. there's already some assumption/expectation that if some outside caller is calling the Hats address, you know exactly what it is and therefore know what functions you can call on it.

but idk. i think it would be cleaner if they implemented ERC165

test/HatsProposalCreationWhitelistV1.test.ts
adamgall: consequence: the strategy is broken from the get-go
solution: deploy a new strategy and configure it correctly

again since this `setUp` call is a one time thing, and our own app is the one calling it, I'm not too concerned about this edge case ever happening in real life.

test/HatsProposalCreationWhitelistV1.test.ts
adamgall: I can add a test for this

contracts/azorius/strategies/HatsProposalCreationWhitelistV1.sol
adamgall: however, I will change the error name from `InvalidHatsContract` to `MissingHatsContract` which better reflects the intention of this zero address check.

adamgall:
@twodam i've implemented the changes i said i would, would appreciate a re-review when you get the chance. thank you!

contracts/azorius/strategies/HatsProposalCreationWhitelistV1.sol
```diff
@@ -0,0 +1,115 @@
...
+     * is wearing any of them using the Hats Protocol.
+     */
+    function isProposer(address _address) public view virtual returns (bool) {
+        for (uint256 i = 0; i < whitelistedHatIds.length; i++) {
+            if (hatsContract.isWearerOfHat(_address, whitelistedHatIds[i])) {
+                return true;
+            }
+        }
+        return false;
+    }
```
dacebt, line 98: should we also be doing other checks here for standing of that hat? For example there are functions like:

https://github.com/Hats-Protocol/hats-protocol/blob/main/src/Hats.sol
- `checkHatWearerStatus` Line 372
- `isActive` Line 1042
- `isInGoodStanding` Line 1066

Or wearer being all we are looking for in this system?

adamgall: - `checkHatWearerStatus` is not a view function, it can update state. so I don't want to call it in this `isProposer` function, which must be a view function
- `isWearerOfHat` calls `balanceOf` which checks both `isActive` and `isEligible`
- `isInGoodStanding` is interesting... I'm honestly not sure about this. Looking at the implementation, it calls a function on the given hat's `eligibility` module. We don't use `eligibility` modules in our system (afaik), so I don't think this is really relevant for us. I've only put a few minutes of thought into this at this point though.

dacebt: - makes sense.

For the other two. yeah. I'm down to talk about this further.

twodam (approved):

dacebt (approved):

PR #221, opened 2025-04-08
feat: add LightAccount verification system

adamgall:
Depends on https://github.com/decentdao/decent-contracts/pull/220

feat: add LightAccount verification system

Add a new verification system for Light Accounts that ensures accounts are created through the official factory contract. This system provides a secure way to validate Light Account authenticity in other contracts.

Key components:
- New abstract `SmartAccountVerificationV1` contract with core verification logic
- `ILightAccountFactory` interface defining the factory contract's required methods
- Comprehensive test suite with mock contracts for testing various scenarios

Technical details:
- Verification checks both contract existence and proper factory deployment
- Uses try/catch pattern to safely handle non-compliant contracts
- Assumes single account per user (salt=0) for Decent App compatibility
- Implements OpenZeppelin's Initializable pattern for proxy support

The verification process:
1. Checks if target is a contract (has code)
2. Verifies implementation of owner() function
3. Validates account address matches factory-derived address
4. Handles all edge cases (EOAs, invalid contracts, wrong factory)

New files:
- contracts/account-abstraction/SmartAccountVerificationV1.sol
- contracts/interfaces/ILightAccountFactory.sol
- contracts/mocks/ConcreteSmartAccountVerification.sol
- contracts/mocks/MockInvalidLightAccount.sol
- contracts/mocks/MockLightAccountFactory.sol
- test/account-abstraction/SmartAccountVerificationV1.test.ts

### DIFF
--- a/contracts/mocks/MockLightAccountFactory.sol
+++ b/contracts/mocks/MockLightAccountFactory.sol
@@ -1,0 +1,23 @@
+// SPDX-License-Identifier: AGPL-3.0
+pragma solidity ^0.8.19;
+
+import {ILightAccountFactory} from "../interfaces/ILightAccountFactory.sol";
+
+contract MockLightAccountFactory is ILightAccountFactory {
+    mapping(address => mapping(uint256 => address)) private _accountAddresses;
+
+    function setAccountAddress(
+        address owner,
+        uint256 salt,
+        address account
+    ) external {
+        _accountAddresses[owner][salt] = account;
+    }
+
+    function getAddress(
+        address owner,
+        uint256 salt
+    ) external view override returns (address) {
+        return _accountAddresses[owner][salt];
+    }
+}
--- a/contracts/account-abstraction/SmartAccountVerificationV1.sol
+++ b/contracts/account-abstraction/SmartAccountVerificationV1.sol
@@ -1,0 +1,54 @@
+// SPDX-License-Identifier: AGPL-3.0
+pragma solidity ^0.8.19;
+
+import {ILightAccount} from "../interfaces/ILightAccount.sol";
+import {ILightAccountFactory} from "../interfaces/ILightAccountFactory.sol";
+import {Initializable} from "@openzeppelin/contracts-upgradeable/proxy/utils/Initializable.sol";
+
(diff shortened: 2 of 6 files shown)


johnqh (approved):

contracts/mocks/MockLightAccountFactory.sol
```diff
@@ -0,0 +1,23 @@
+// SPDX-License-Identifier: AGPL-3.0
+pragma solidity ^0.8.19;
+
+import {ILightAccountFactory} from "../interfaces/ILightAccountFactory.sol";
+
+contract MockLightAccountFactory is ILightAccountFactory {
+    mapping(address => mapping(uint256 => address)) private _accountAddresses;
+
+    function setAccountAddress(
```
DarksightKellar, line 9: You could have fun with this and name it `deployAccount` or `createAccount` to more closely match the actual implementation signature

test/account-abstraction/LightAccountVerificationV1.test.ts
```diff
@@ -0,0 +1,135 @@
...
+          void expect(await concreteVerification.verifyLightAccountPublic(randomAddress)).to.be
+            .false;
+        });
+      });
+
+      describe('contracts', function () {
+        it('should return false for invalid light accounts that do implement the ILightAccount interface (owner())', async function () {
+          // not calling "setAccountAddress", so the LightAccountFactory will always
+          // return the zero address when calling `getAddress` for a given owner and salt
+          // (which is implemented in the LightAccountVerification verifyLightAccount function).
```
DarksightKellar: Then this comment would be more intuitive, because the test's (mock) light account factory hasn't "deployed" the `mockLightAccount`

DarksightKellar (approved):
Approving, but left a name suggestion in there

contracts/mocks/MockLightAccountFactory.sol
adamgall: I get where you're coming from, but I'd prefer to leave it as-is, because the current name is most representative of what's actually happening in the mock factory contract. nothing is being deployed or created.

PR #238, opened 2025-04-14
chore: remove IProxyFactory interface and update prepublish script for correct network handling

adamgall:
- This commit deletes the IProxyFactory interface from the contracts/interfaces directory, which should have happened in this PR: https://github.com/decentdao/decent-contracts/pull/218.
- Additionally, the prepublish script is updated to extract ABI values from the Sepolia network key in the deployments.json file, ensuring that release candidate npm publishes capture "all" contracts, assuming that contracts are being deployed to Sepolia first.


adamgall:
not going to wait for reviews for this one.

PR #248, opened 2025-05-08
Feature: Add LinearERC20VotingV1ValidatorV1

adamgall:
Depends on https://github.com/decentdao/decent-contracts/pull/254

Fixes [ENG-818](https://linear.app/decent-labs/issue/ENG-818/develop-validator-for-linearerc20votingv1)

**Summary:**

This PR introduces `LinearERC20VotingV1ValidatorV1`, a dedicated validator contract designed to work with `DecentPaymasterV1`. This validator specifically checks if a `vote(uint32, uint8)` UserOperation targeting a `LinearERC20VotingV1` strategy contract is likely to succeed before the paymaster agrees to sponsor it.

The PR includes the validator contract itself, a mock version of `LinearERC20VotingV1` for testing, and a comprehensive test suite for the new validator.

**Key Components:**

1.  **New Contract `LinearERC20VotingV1ValidatorV1.sol`:**
    *   Implements `IFunctionValidator`, `ERC165`, and `Version`.
    *   **`validateOperation(...)`:**
        *   Verifies the `callData` selector matches `ILinearERC20VotingV1.vote.selector`.
        *   Decodes `proposalId` and `voteType` from `callData`.
        *   Checks for valid `voteType` (0, 1, or 2).
        *   Retrieves proposal period (`startBlock`, `endBlock`) from the target `votingContract`.
        *   Ensures the proposal exists (non-zero `endBlock`).
        *   Verifies the voting period has not ended.
        *   Checks if the `lightAccountOwner` has already voted on the proposal.
        *   Retrieves the `governanceToken` from the `votingContract`.
        *   Fetches the `lightAccountOwner`'s checkpoints from the `governanceToken`.
        *   Returns `false` if the voter has no checkpoints.
        *   Iterates through checkpoints to find the voting weight at the `proposal.startBlock`.
        *   Returns `false` if the voting weight at `startBlock` is zero.
        *   Returns `true` if all checks pass, indicating the vote operation is likely to succeed.
    *   **`getVersion()`:** Returns the contract version (1).
    *   **`supportsInterface(...)`:** Implements ERC165 support for `IFunctionValidator`, `IVersion`, and `IERC165`.

2.  **New Mock Contract `MockLinearERC20VotingV1.sol`:**
    *   Provides a mock implementation of the functions from `LinearERC20VotingV1` and `IERC20Votes` that the validator interacts with.
    *   Allows setting proposal periods, voting ended status, voter status (`hasVoted`), governance token address, and checkpoint data for test scenarios.

3.  **New Test Suite `LinearERC20VotingV1ValidatorV1.test.ts`:**
    *   Provides comprehensive test coverage for `LinearERC20VotingV1ValidatorV1`.
    *   Tests various scenarios for `validateOperation`, including:
        *   Incorrect function selector in `callData`.
        *   Invalid vote types.
        *   Non-existent proposals.
        *   Voting period ended.
        *   User has already voted.
        *   Voter has zero voting weight (no checkpoints, all checkpoints after proposal start, relevant checkpoint has zero votes).
        *   Valid vote operations with various single and multiple checkpoint configurations.
        *   Edge cases like max voting weight and proposal start at block 0.
    *   Includes tests for ERC165 interface support and versioning.

**Motivation:**

*   To enable gasless voting for users interacting with `LinearERC20VotingV1` strategies via `DecentPaymasterV1`.
*   To provide a robust pre-validation mechanism that ensures the paymaster only sponsors vote operations that are likely to succeed, saving gas and preventing unnecessary sponsored transactions.
*   To create a reusable and upgradeable validator specific to the `LinearERC20VotingV1` contract's voting logic.

### DIFF
--- a/contracts/deployables/account-abstraction/validators/LinearERC20VotingV1ValidatorV1.sol
+++ b/contracts/deployables/account-abstraction/validators/LinearERC20VotingV1ValidatorV1.sol
@@ -1,0 +1,179 @@
+// SPDX-License-Identifier: AGPL-3.0
+pragma solidity ^0.8.28;
+
+import {IFunctionValidator} from "../../../interfaces/decent/deployables/IFunctionValidator.sol";
+import {Version} from "../../Version.sol";
+import {ERC165} from "@openzeppelin/contracts/utils/introspection/ERC165.sol";
+import {IERC165} from "@openzeppelin/contracts/utils/introspection/IERC165.sol";
+import {UUPSUpgradeable} from "@openzeppelin/contracts-upgradeable/proxy/utils/UUPSUpgradeable.sol";
+import {OwnableUpgradeable} from "@openzeppelin/contracts-upgradeable/access/OwnableUpgradeable.sol";
+
+interface ILinearERC20VotingV1 {
+    function vote(uint32 proposalId, uint8 voteType) external;
+
+    function hasVoted(
+        uint32 proposalId,
+        address account
+    ) external view returns (bool);
+
+    function getProposalPeriod(
+        uint32 proposalId
+    ) external view returns (uint32, uint32);
+
+    function votingPeriodEnded(uint32 proposalId) external view returns (bool);
+
+    function governanceToken() external view returns (address);
+}
+
+struct Checkpoint {
+    uint32 fromBlock;
+    uint224 votes;
+}
+
+interface IERC20Votes {
+    function numCheckpoints(address account) external view returns (uint32);
+
+    function checkpoints(
+        address account,
+        uint32 pos
+    ) external view returns (Checkpoint memory);
+}
+
+/**
+ * @title LinearERC20VotingV1ValidatorV1
+ * @dev Validates vote operations for LinearERC20VotingV1 to ensure they will succeed
+ */
+contract LinearERC20VotingV1ValidatorV1 is
+    IFunctionValidator,
+    ERC165,
+    Version,
+    UUPSUpgradeable,
+    OwnableUpgradeable
+{
+    uint16 public constant VERSION = 1;
+
+    function initialize(address owner) public initializer {
+        __Ownable_init(owner);
+    }
+
+    function _authorizeUpgrade(
+        address newImplementation
+    ) internal override onlyOwner {}
+
+    /**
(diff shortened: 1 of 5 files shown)


dacebt (commented):
I'm stuck on these validators. Where are these used deployed/setup/used in the Gasless voting setup

dacebt (approved):
Thanks for the InG update, makes sense. Curious just had a thought this is seems very specific to voting and you said that only one validator will be assigned. Will this pattern allow for gasless options in the future?

adamgall:
> Thanks for the InG update, makes sense. Curious just had a thought this is seems very specific to voting and you said that only one validator will be assigned. Will this pattern allow for gasless options in the future?

One validator is assigned _per target contract and function_. So yes, this pattern allows for additional transaction sponsorship (more than just voting) in the future.

contracts/deployables/account-abstraction/validators/LinearERC20VotingV1ValidatorV1.sol
```diff
@@ -0,0 +1,163 @@
...
+        // Iterate backwards through checkpoints to find the relevant one for startBlock.
+        // This is potentially more efficient than binary search if startBlock is recent.
+        uint256 votingWeight = 0;
+        for (uint256 i = numCheckpoints; i > 0; i--) {
+            // Checkpoint indices are 0-based, loop index 'i' is 1-based count.
+            Checkpoint memory checkpoint = governanceToken.checkpoints(
+                lightAccountOwner,
+                uint32(i - 1)
+            );
+
```
johnqh, line 128: I was thinking about this and there is another optimization we can do:
if (checkpoint.fromBlock > endBlock) {
// This means the vote has ended, since there is a checkpoint after the proposal endBlock. We can break out of the loop and return false.
}

johnqh (changes requested):
Additional optimization we can do

contracts/deployables/account-abstraction/validators/LinearERC20VotingV1ValidatorV1.sol
adamgall: updated

johnqh (approved):

PR #257, opened 2025-05-14
refactor(FreezeVoting): Transition freeze voting to time-based periods

adamgall:
Depends on https://github.com/decentdao/decent-contracts/pull/256

Fixes [ENG-849](https://linear.app/decent-labs/issue/ENG-849/refactor-all-freeze-voting-contracts-to-use-time-based-periods-seconds)

Converts `BaseFreezeVotingV1` and its concrete implementations (`ERC20FreezeVotingV1`, `ERC721FreezeVotingV1`, `MultisigFreezeVotingV1`) from using block numbers to using timestamps for freeze proposal and duration periods.

Key changes include:
- `freezeProposalCreatedBlock` (uint32) is renamed to `freezeProposalCreated` and now stores `block.timestamp` (as `uint48`).
- `freezeProposalPeriod` and `freezePeriod` now represent durations in seconds, and their descriptions are updated accordingly.
- The `userHasFreezeVoted` mapping in `BaseFreezeVotingV1` now uses the `freezeProposalCreated` timestamp as its second key.
- The `idHasFreezeVoted` mapping in `ERC721FreezeVotingV1` also now uses the `freezeProposalCreated` timestamp as its first key.
- All internal logic, including checking for proposal expiry and freeze duration in `isFrozen()`, now uses `block.timestamp`.
- Tests for all affected freeze voting contracts and the base contract are updated to use Hardhat Network Helpers' `time` functions (e.g., `time.increase`, `time.latest`) instead of `mine`.
- Interface descriptions and mock implementations are updated to reflect the change from blocks to seconds/timestamps.

This refactoring ensures that the timing of freeze proposals and the duration of freezes are determined by actual time, making the system more consistent and less dependent on block production rates.

### DIFF
--- a/contracts/deployables/freeze-voting/ERC20FreezeVotingV1.sol
+++ b/contracts/deployables/freeze-voting/ERC20FreezeVotingV1.sol
@@ -32,8 +32,8 @@ contract ERC20FreezeVotingV1 is BaseFreezeVotingV1, Version {
      *
      * @param _owner The owner of the contract
      * @param _freezeVotesThreshold The number of votes required to activate a freeze
-     * @param _freezeProposalPeriod The number of blocks a freeze proposal has to succeed
-     * @param _freezePeriod The number of blocks a freeze lasts
+     * @param _freezeProposalPeriod The number of seconds a freeze proposal has to succeed
+     * @param _freezePeriod The number of seconds a freeze lasts
      * @param _votesERC20 The ERC20 voting token contract address
      */
     function initialize(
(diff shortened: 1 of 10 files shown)


twodam (approved):

contracts/deployables/freeze-voting/ERC20FreezeVotingV1.sol
```diff
@@ -65,14 +65,14 @@ contract ERC20FreezeVotingV1 is BaseFreezeVotingV1, Version {
...
+        if (block.timestamp > freezeProposalCreated + freezeProposalPeriod) {
             // create a new freeze proposal and set total votes to msg.sender's vote count
 
-            freezeProposalCreatedBlock = uint32(block.number);
+            freezeProposalCreated = uint48(block.timestamp);
 
             userVotes = votesERC20.getPastVotes(
                 msg.sender,
-                freezeProposalCreatedBlock - 1
+                freezeProposalCreated - 1
```
dacebt, line 75: Doesn't `getPastVotes` take a block.number as the second param? would this still work passing in a timestamp?

adamgall: it's dependent on whatever the `clock()` on the ERC20 contract returns: a block number or a block timestamp.

if the clock is using block numbers, then the value into `getPastVotes` needs to be a block number.

if the clock is using timestamps, then the value into `getPastVotes` needs to be a timestamp.

dacebt: Ah okay, yep yep `EIP-6372` got it.

dacebt (approved):

johnqh (approved):

PR #258, opened 2025-05-15
feat(governance): Introduce ClockModeLib for standardized time handling

adamgall:
Depends on https://github.com/decentdao/decent-contracts/pull/257

Fixes [ENG-855](https://linear.app/decent-labs/issue/ENG-855/implement-clockmodelib-for-standardized-time-point-resolution)

This commit introduces a `ClockMode` enum and `ClockModeLib` to standardize how the system determines whether to use `block.timestamp` or `block.number` for time-sensitive operations, particularly when interacting with tokens that may implement ERC6372.

The `ClockModeLib` provides:
- `getClockMode(address token)`: Detects the clock mode of a given token. It defaults to `BlockNumber` if the token does not implement `CLOCK_MODE()` as per ERC6372, if `CLOCK_MODE()` reverts, or if it returns a non-standard string.
- `getCurrentPoint(ClockMode mode)`: Returns either `block.timestamp` or `block.number` based on the specified `ClockMode`.

Includes:
- `ClockMode.sol`: Enum for `Timestamp` and `BlockNumber` modes.
- `ClockModeLib.sol`: Library with detection and utility functions.
- Mock contracts (`MockTimestampToken`, `MockBlockNumberToken`, `MockRevertingToken`, `MockNonIERC6372Token`) for thorough testing.
- `ClockModeLibTester.sol`: A contract to facilitate testing the library.
- Comprehensive Hardhat tests for `ClockModeLib`.

### DIFF
--- a/contracts/interfaces/decent/ClockMode.sol
+++ b/contracts/interfaces/decent/ClockMode.sol
@@ -1,0 +1,11 @@
+// SPDX-License-Identifier: AGPL-3.0
+pragma solidity ^0.8.30;
+
+/**
+ * @title ClockMode
+ * @dev Enum to distinguish between timestamp-based and block number-based time.
+ */
+enum ClockMode {
+    Timestamp,
+    BlockNumber
+}
--- a/contracts/libs/ClockModeLib.sol
+++ b/contracts/libs/ClockModeLib.sol
@@ -1,0 +1,45 @@
+// SPDX-License-Identifier: AGPL-3.0
+pragma solidity ^0.8.30;
+
+import {ClockMode} from "../interfaces/decent/ClockMode.sol";
+import {IERC6372} from "@openzeppelin/contracts/interfaces/IERC6372.sol";
+
+/**
+ * @title ClockModeLib
+ * @dev Library for handling operations related to ClockMode.
+ */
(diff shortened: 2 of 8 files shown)


dacebt (commented):
Interesting so would we need to do this check in the frontend as well for our 'import token' feature? It could be `IVotes` but the wrong ClockMode?

I guess another question I have is the clock mode of a token fixed or dynamic?

johnqh (approved):

adamgall:
> Interesting so would we need to do this check in the frontend as well for our 'import token' feature? It could be `IVotes` but the wrong ClockMode?

So yeah, _right now_ in the frontend, we _should_ be checking that the clock mode is block number based, and if not, disallow. If it's timestamp based, all of our proposal actions (voting) and time calculations (how long till proposal is closed, etc) will be wrong.

The next thing to do in the frontend, _right now_ (potentially before these contracts go live), is to adjust the frontend to properly handle `clock()`s that are timestamp based.

Then when these new contracts go live and start being used, we'll already be good to go.

> 
> I guess another question I have is the clock mode of a token fixed or dynamic?

We can assume it'll be fixed.

dacebt (approved):

PR #266, opened 2025-05-16
Feature/counter sign

johnqh:
I have talked to various people about a CounterSign contract. I was asked to create some design doc, but I thought code is easier to read and understand.

Nowhere to be final, but I think you can get the idea

Also created a use case walkthrough at https://docs.google.com/document/d/1xp8w25O7CnsMf3cVPX_AZ3XT0k6DaAaWWl50p-VCj9A/edit?usp=sharing

### DIFF
--- a/contracts/deployables/counter-sign/CounterSignV1.sol
+++ b/contracts/deployables/counter-sign/CounterSignV1.sol
@@ -16,7 +16,6 @@ contract CounterSignV1 is Initializable, ICounterSignV1 {
     address[] private signers;
     mapping(address => string) private entityNames;
     mapping(address => string) private signerPersonalNames;
-    mapping(address => Transaction[]) private signerTransactions;
     mapping(address => bool) public signedAddresses;
 
     Transaction[] private finalAgreementTransactions;
@@ -30,29 +29,17 @@ contract CounterSignV1 is Initializable, ICounterSignV1 {
         (
             address[] memory _signers,
             string[] memory _entityNames,
-            Transaction[][] memory _signerTxs,
             uint256 _expiration,
             string memory _agreementUrl,
             Transaction[] memory _finalTxs
         ) = abi.decode(
                 params,
-                (
-                    address[],
-                    string[],
-                    Transaction[][],
-                    uint256,
-                    string,
-                    Transaction[]
-                )
+                (address[], string[], uint256, string, Transaction[])
             );
 
         require(
             _signers.length == _entityNames.length,
             "Signers and names mismatch"
-        );
-        require(
-            _signers.length == _signerTxs.length,
-            "Signers and transactions mismatch"
         );
 
         signers = _signers;
@@ -65,11 +52,6 @@ contract CounterSignV1 is Initializable, ICounterSignV1 {
         for (uint256 i = 0; i < _signers.length; i++) {
             address signer = _signers[i];
             entityNames[signer] = _entityNames[i];
-
-            Transaction[] memory txs = _signerTxs[i];
-            for (uint256 j = 0; j < txs.length; j++) {
-                signerTransactions[signer].push(txs[j]);
-            }
         }
     }
 
@@ -85,11 +67,6 @@ contract CounterSignV1 is Initializable, ICounterSignV1 {
 
         signedAddresses[msg.sender] = true;
         signerPersonalNames[msg.sender] = name;
-
-        Transaction[] storage txs = signerTransactions[msg.sender];
-        for (uint256 i = 0; i < txs.length; i++) {
-            _executeTransaction(txs[i]);
-        }
     }
 
     function executeAgreement() external override {
--- a/contracts/deployables/counter-sign/CounterSignV1.sol
+++ b/contracts/deployables/counter-sign/CounterSignV1.sol
@@ -1,0 +1,152 @@
+// SPDX-License-Identifier: AGPL-3.0
+pragma solidity ^0.8.30;
+
+import {ICounterSignV1} from "../../interfaces/decent/deployables/ICounterSignV1.sol";
+import {Initializable} from "@openzeppelin/contracts-upgradeable/proxy/utils/Initializable.sol";
+
+contract CounterSignV1 is Initializable, ICounterSignV1 {
+    string private agreementUrl;
+    uint256 private expirationTimestamp;
+    bool private executed;
+
+    address[] private signers;
+    mapping(address => string) private entityNames;
+    mapping(address => string) private signerPersonalNames;
+    mapping(address => Transaction[]) private signerTransactions;
+    mapping(address => bool) public signedAddresses;
+
+    Transaction[] private finalAgreementTransactions;
+
+    modifier onlyBeforeExpiry() {
+        require(block.timestamp < expirationTimestamp, "Agreement expired");
+        _;
+    }
+
+    function initialize(bytes memory params) external initializer {
+        (
+            address[] memory _signers,
+            string[] memory _entityNames,
+            Transaction[][] memory _signerTxs,
+            uint256 _expiration,
+            string memory _agreementUrl,
+            Transaction[] memory _finalTxs
+        ) = abi.decode(
+                params,
+                (
+                    address[],
+                    string[],
+                    Transaction[][],
+                    uint256,
+                    string,
+                    Transaction[]
+                )
+            );
+
+        require(
+            _signers.length == _entityNames.length,
+            "Signers and names mismatch"
+        );
+        require(
+            _signers.length == _signerTxs.length,
+            "Signers and transactions mismatch"
+        );
+
+        signers = _signers;
+        expirationTimestamp = _expiration;
+        agreementUrl = _agreementUrl;
+        for (uint256 i = 0; i < _finalTxs.length; i++) {
+            finalAgreementTransactions.push(_finalTxs[i]);
+        }
+
+        for (uint256 i = 0; i < _signers.length; i++) {
+            address signer = _signers[i];
+            entityNames[signer] = _entityNames[i];
+
+            Transaction[] memory txs = _signerTxs[i];
+            for (uint256 j = 0; j < txs.length; j++) {
+                signerTransactions[signer].push(txs[j]);
+            }
+        }
+    }
+
+    function getAgreementUrl() external view override returns (string memory) {
+        return agreementUrl;
+    }
+
+    function signAgreement(
+        string memory name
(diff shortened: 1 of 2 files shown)


contracts/deployables/counter-sign/CounterSignV1.sol
```diff
@@ -0,0 +1,133 @@
...
+        (
+            address[] memory _signers,
+            string[] memory _entityNames,
+            uint256 _expiration,
+            string memory _agreementUrl,
+            Transaction[] memory _finalTxs
+        ) = abi.decode(
+                params,
+                (address[], string[], uint256, string, Transaction[])
+            );
```
adamgall, line 38: You don't need to use the `bytes memory params` pattern here, assuming we deploy instances of this through the new `ProxyFactory` contract. You can just include the params as normal parameters to the `initialize` function.

contracts/deployables/counter-sign/CounterSignV1.sol
```diff
@@ -0,0 +1,133 @@
...
+
+        require(
+            _signers.length == _entityNames.length,
+            "Signers and names mismatch"
+        );
```
adamgall, line 43: nit: use proper solidity `error` types instead of `require`.

```sol
error InputMismatch();

function initialize(...) {
  ...
  if (_signers.length != _entityNames.length) {
    revert InputMismatch();
  }
}
```

contracts/deployables/counter-sign/CounterSignV1.sol
```diff
@@ -0,0 +1,133 @@
...
+
+import {ICounterSignV1} from "../../interfaces/decent/deployables/ICounterSignV1.sol";
+import {Initializable} from "@openzeppelin/contracts-upgradeable/proxy/utils/Initializable.sol";
+
+/*
+Design discussion: https://docs.google.com/document/d/1xp8w25O7CnsMf3cVPX_AZ3XT0k6DaAaWWl50p-VCj9A/edit?usp=sharing
+*/
+
+contract CounterSignV1 is Initializable, ICounterSignV1 {
+    string private agreementUrl;
```
adamgall, line 12: nit: rename to `agreementUri` to be more inclusive of different ways we might store this data.

For example, an IPFS identifier can be defined as `ipfs://<cid>`, which does identify a specific IPFS document but does not specify the gateway used to retrieve it (leaving that up to the client). In this case, "`ipfs://<cid>`" is technically a "URI", not a "URL".

contracts/deployables/counter-sign/CounterSignV1.sol
```diff
@@ -0,0 +1,133 @@
...
+
+    address[] private signers;
+    mapping(address => string) private entityNames;
+    mapping(address => string) private signerPersonalNames;
+    mapping(address => bool) public signedAddresses;
+
+    Transaction[] private finalAgreementTransactions;
+
+    modifier onlyBeforeExpiry() {
+        require(block.timestamp < expirationTimestamp, "Agreement expired");
```
adamgall, line 24: nit: use error types here.

contracts/deployables/counter-sign/CounterSignV1.sol
```diff
@@ -0,0 +1,133 @@
...
+        signers = _signers;
+        expirationTimestamp = _expiration;
+        agreementUrl = _agreementUrl;
+        for (uint256 i = 0; i < _finalTxs.length; i++) {
+            finalAgreementTransactions.push(_finalTxs[i]);
+        }
+
+        for (uint256 i = 0; i < _signers.length; i++) {
+            address signer = _signers[i];
+            entityNames[signer] = _entityNames[i];
```
adamgall, line 54: would be less gas to not create `signer` variable here:

```sol
entityNames[_signers[i]] = _entityNames[i];
```

contracts/deployables/counter-sign/CounterSignV1.sol
```diff
@@ -0,0 +1,133 @@
...
+            _signers.length == _entityNames.length,
+            "Signers and names mismatch"
+        );
+
+        signers = _signers;
+        expirationTimestamp = _expiration;
+        agreementUrl = _agreementUrl;
+        for (uint256 i = 0; i < _finalTxs.length; i++) {
+            finalAgreementTransactions.push(_finalTxs[i]);
+        }
```
adamgall, line 50: why the loop?

```sol
finalAgreementTransactions = _finalTxs;
```

contracts/deployables/counter-sign/CounterSignV1.sol
```diff
@@ -0,0 +1,133 @@
...
+        bool success;
+        if (txn.operation == Operation.Call) {
+            (success, ) = txn.to.call{value: txn.value}(txn.data);
+        } else if (txn.operation == Operation.DelegateCall) {
+            (success, ) = txn.to.delegatecall(txn.data);
+        }
+        require(success, "Transaction execution failed");
+    }
+
+    receive() external payable {}
```
adamgall, line 132: I assume the intention here is that this contract might need to receive ETH?

Considering we said on the call earlier that this contract shouldn't act as escrow, what is the use case for this contract receiving ETH?

I think this should be removed.

contracts/deployables/counter-sign/CounterSignV1.sol
```diff
@@ -0,0 +1,133 @@
...
+
+    function _isSigner(address addr) internal view returns (bool) {
+        for (uint256 i = 0; i < signers.length; i++) {
+            if (signers[i] == addr) return true;
+        }
+        return false;
+    }
+
+    function _executeTransaction(Transaction memory txn) internal {
+        bool success;
```
adamgall, line 123: I know its redundant (because booleans are initialized as "false" already), but I think it'd be nice to see this as `bool success = false`, considering the "if..else if.." below. If neither of those conditions hit, then we are relying on the `require(success, ...)` to revert the transaction, and so from a readability perspective the "default value of success" needs to be false.

contracts/deployables/counter-sign/CounterSignV1.sol
```diff
@@ -0,0 +1,133 @@
...
+
+    function executeAgreement() external override {
+        require(getIsExecutable(), "Not all signed");
+        require(!executed, "Already executed");
+
+        for (uint256 i = 0; i < finalAgreementTransactions.length; i++) {
+            _executeTransaction(finalAgreementTransactions[i]);
+        }
+        executed = true;
+    }
```
adamgall, line 80: Emit an event in this function.

```sol
emit AgreementExecuted();
```

contracts/deployables/counter-sign/CounterSignV1.sol
```diff
@@ -0,0 +1,133 @@
...
+
+    function signAgreement(
+        string memory name
+    ) external override onlyBeforeExpiry {
+        require(_isSigner(msg.sender), "Not a signer");
+        require(!signedAddresses[msg.sender], "Already signed");
+
+        signedAddresses[msg.sender] = true;
+        signerPersonalNames[msg.sender] = name;
+    }
```
adamgall, line 70: Emit an event in this function.

```sol
emit AgreementSigned(msg.sender, name);
```

contracts/deployables/counter-sign/CounterSignV1.sol
```diff
@@ -0,0 +1,133 @@
...
+contract CounterSignV1 is Initializable, ICounterSignV1 {
+    string private agreementUrl;
+    uint256 private expirationTimestamp;
+    bool private executed;
+
+    address[] private signers;
+    mapping(address => string) private entityNames;
+    mapping(address => string) private signerPersonalNames;
+    mapping(address => bool) public signedAddresses;
+
```
adamgall, line 20: This comment and the next few all have to do with removing the for loops in `getIsExecutable` and `_isSigner`, which could get expensive if there are many signers.

Start with introducing two new variables:

```sol
mapping(address => bool) private isSignerAddress;
uint256 private signedCount;
```

contracts/deployables/counter-sign/CounterSignV1.sol
```diff
@@ -0,0 +1,133 @@
...
+        expirationTimestamp = _expiration;
+        agreementUrl = _agreementUrl;
+        for (uint256 i = 0; i < _finalTxs.length; i++) {
+            finalAgreementTransactions.push(_finalTxs[i]);
+        }
+
+        for (uint256 i = 0; i < _signers.length; i++) {
+            address signer = _signers[i];
+            entityNames[signer] = _entityNames[i];
+        }
```
adamgall, line 55: In this for loop, do `isSignerAddress[signer] = true;`

(see https://github.com/decentdao/decent-contracts/pull/266#discussion_r2093656892)

contracts/deployables/counter-sign/CounterSignV1.sol
adamgall, line 70: At the end of this function, do `signedCount++;`

(see https://github.com/decentdao/decent-contracts/pull/266#discussion_r2093656892)

contracts/deployables/counter-sign/CounterSignV1.sol
```diff
@@ -0,0 +1,133 @@
...
+    ) external view override returns (bool) {
+        return signedAddresses[signerAddress];
+    }
+
+    function getIsExecutable() public view override returns (bool) {
+        for (uint256 i = 0; i < signers.length; i++) {
+            if (!signedAddresses[signers[i]]) return false;
+        }
+        return true;
+    }
```
adamgall, line 97: Replace this whole function with simply `return signedCount == signers.length;`

(see https://github.com/decentdao/decent-contracts/pull/266#discussion_r2093656892)

contracts/deployables/counter-sign/CounterSignV1.sol
```diff
@@ -0,0 +1,133 @@
...
+    ) external view returns (string memory) {
+        return signerPersonalNames[signerAddress];
+    }
+
+    function _isSigner(address addr) internal view returns (bool) {
+        for (uint256 i = 0; i < signers.length; i++) {
+            if (signers[i] == addr) return true;
+        }
+        return false;
+    }
```
adamgall, line 120: Replace this whole function with simply `return isSignerAddress[addr];`

(see https://github.com/decentdao/decent-contracts/pull/266#discussion_r2093656892)

adamgall (changes requested):

adamgall:
Closing because @tbwebb22 is working on this in another set of branches.

PR #269, opened 2025-05-23
refactor(AzoriusV1)!: Transition to Single Strategy Model

adamgall:
Fixes [ENG-929](https://linear.app/decent-labs/issue/ENG-929/refactor-azoriusv1-to-support-a-single-pluggable-istrategyv1)

**High-Level Context:**

This Pull Request is a pivotal step in a broader architectural refactoring of our governance system. We are moving away from `AzoriusV1` managing multiple, distinct voting strategy contracts towards a model where `AzoriusV1` interacts with a single, more powerful and flexible `IStrategyV1` compliant contract. This new strategy (to be detailed in subsequent PRs/related issues, like `StrategyV1.sol` with its token adapter pattern) will encapsulate more complex voting logic.

This specific PR focuses on modifying `AzoriusV1` and its interface `IAzoriusV1` to support this single-strategy architecture, simplifying `AzoriusV1`'s responsibilities.

**Changes in this PR:**

1.  **Single Strategy Configuration:**
    *   Removed the internal linked-list mechanism for managing multiple strategies (`SENTINEL_STRATEGY`, `strategies` mapping).
    *   `AzoriusV1` now holds a reference to a single `IStrategyV1 public strategy`.
    *   The `IBaseStrategyV1` import has been updated to `IStrategyBaseV1`.

2.  **Initialization and Setup:**
    *   The `initialize` and `setUp` methods have been updated to accept a single `_strategy` address instead of an array of strategies.
    *   The internal `_setUpStrategies` function has been removed.
    *   A new internal function `_updateStrategy(address _strategy)` and an `external onlyOwner` function `updateStrategy(address _strategy)` have been added to set and update the single strategy, emitting a `StrategyUpdated` event.

3.  **Proposal Submission:**
    *   `submitProposal` no longer accepts `_strategy` or `_data` parameters.
    *   It now directly uses the configured `strategy` and calls `strategy.initializeProposal(totalProposalCount)`. The `_data` parameter previously passed to the strategy is removed at this level; if the new `IStrategyV1` needs data, it will be handled differently (e.g., not passed via Azorius `submitProposal`).
    *   The `ProposalCreated` event no longer emits the strategy address, as it's fixed for the Azorius instance.

4.  **Strategy Management Functions Removed:**
    *   Functions related to managing multiple strategies like `getStrategies`, `enableStrategy`, `disableStrategy`, and `isStrategyEnabled` have been removed.
    *   Associated events (`EnabledStrategy`, `DisabledStrategy`) and errors (`InvalidStrategy`, `InvalidStartAddress`, `InvalidCount`, `StrategyEnabled`, `StrategyDisabled`) have also been removed.

5.  **Proposal Data and State:**
    *   The `Proposal` struct in `AzoriusV1` still stores `strategy` (now the address of the *single* strategy active at the time of proposal creation, which is `address(this.strategy)`).
    *   `getProposal` continues to return the strategy address associated with a specific proposal.
    *   `proposalState` now uses the globally configured `this.strategy` to determine proposal outcomes.

**Impact:**
*   **Breaking Change:** This PR introduces breaking changes to the `AzoriusV1` contract interface and its initialization parameters.
*   **Simplification:** `AzoriusV1` becomes simpler by delegating all strategy-specific logic and management to a single, swappable strategy contract.
*   **Foundation for Modularity:** This change is crucial for the upcoming introduction of `StrategyV1` and its Token Adapter pattern, which will provide enhanced flexibility and extensibility for defining voting mechanisms.

**Current Status:**
*   The changes in this PR make `AzoriusV1` ready to work with an `IStrategyV1` compliant contract.
*   **Compilation of `AzoriusV1` itself should succeed with these changes, assuming a valid `IStrategyV1` interface is present.**
*   **However, the overall project and test suite will still have significant failures.** This is because the concrete strategy contracts (like the old `LinearERC20VotingV1`, `LinearERC721VotingV1`) and other parts of the system have not yet been updated to conform to this new single-strategy model or the new `IStrategyV1` interface they would need to implement. These will be addressed in subsequent PRs.

### DIFF
--- a/contracts/deployables/modules/AzoriusV1.sol
+++ b/contracts/deployables/modules/AzoriusV1.sol
@@ -2,31 +2,13 @@
 pragma solidity ^0.8.30;
 
 import {Version} from "../Version.sol";
-import {IBaseStrategyV1} from "../../interfaces/decent/deployables/IBaseStrategyV1.sol";
+import {IStrategyBaseV1} from "../../interfaces/decent/deployables/IStrategyBaseV1.sol";
 import {IAzoriusV1, Enum} from "../../interfaces/decent/deployables/IAzoriusV1.sol";
 import {GuardableModule} from "@gnosis-guild/zodiac/contracts/core/GuardableModule.sol";
 import {UUPSUpgradeable} from "@openzeppelin/contracts-upgradeable/proxy/utils/UUPSUpgradeable.sol";
 
-/**
- * A Safe module which allows for composable governance.
- * Azorius conforms to the [Zodiac pattern](https://github.com/gnosis/zodiac) for Safe modules.
- *
- * The Azorius contract acts as a central manager of DAO Proposals, maintaining the specifications
- * of the transactions that comprise a Proposal, but notably not the state of voting.
- *
- * All voting details are delegated to [BaseStrategy](./BaseStrategy.md) implementations, of which an Azorius DAO can
- * have any number.
- */
 contract AzoriusV1 is IAzoriusV1, GuardableModule, Version, UUPSUpgradeable {
     uint16 private constant VERSION = 1;
-
-    /**
-     * The sentinel node of the linked list of enabled [BaseStrategies](./BaseStrategy.md).
-     *
-     * See https://en.wikipedia.org/wiki/Sentinel_node.
-     */
-    address internal constant SENTINEL_STRATEGY = address(0x1);
-
     /**
      * ```
      * keccak256(
@@ -53,20 +35,15 @@ contract AzoriusV1 is IAzoriusV1, GuardableModule, Version, UUPSUpgradeable {
     bytes32 public constant TRANSACTION_TYPEHASH =
         0x72e9670a7ee00f5fbf1049b8c38e3f22fab7e9b85029e85cf9412f17fdd5c2ad;
 
-    /** Total number of submitted Proposals. */
     uint32 public totalProposalCount;
 
-    /** Delay between when a Proposal is passed and when it can be executed. */
     uint32 public timelockPeriod;
 
-    /** Time between when timelock ends and the Proposal expires. */
     uint32 public executionPeriod;
 
-    /** Proposals by `proposalId`. */
     mapping(uint256 => Proposal) internal proposals;
 
-    /** A linked list of enabled [BaseStrategies](./BaseStrategy.md). */
-    mapping(address => address) internal strategies;
+    IStrategyBaseV1 public strategy;
 
     event AzoriusSetUp(
         address indexed creator,
@@ -82,16 +59,11 @@ contract AzoriusV1 is IAzoriusV1, GuardableModule, Version, UUPSUpgradeable {
         string metadata
     );
     event ProposalExecuted(uint32 proposalId, bytes32[] txHashes);
-    event EnabledStrategy(address strategy);
-    event DisabledStrategy(address strategy);
     event TimelockPeriodUpdated(uint32 timelockPeriod);
     event ExecutionPeriodUpdated(uint32 executionPeriod);
+    event StrategyUpdated(address strategy);
 
     error InvalidStrategy();
-    error InvalidStartAddress();
-    error InvalidCount();
-    error StrategyEnabled();
-    error StrategyDisabled();
     error InvalidProposal();
     error InvalidProposer();
     error ProposalNotExecutable();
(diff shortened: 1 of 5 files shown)


contracts/interfaces/decent/deployables/IAzoriusV1.sol
```diff
@@ -101,18 +82,20 @@ interface IAzoriusV1 {
...
      *
-     * @param _strategy address of the BaseStrategy implementation which the Proposal will use
-     * @param _data arbitrary data passed to the BaseStrategy implementation. This may not be used by all strategies,
-     * but is included in case future strategy contracts have a need for it
      * @param _transactions array of transactions to propose
      * @param _metadata additional data such as a title/description to submit with the proposal
      */
     function submitProposal(
-        address _strategy,
-        bytes memory _data,
```
adamgall, line 115: Might be prudent to keep this `_data` param, in case a `StrategyV2` in the future needs it. We can of course also have a DAO _update_ their `AzoriusV1` to an `AzoriusV2` if we do decide to need this in the future, but adding it back in now would save that headache.

contracts/deployables/modules/AzoriusV1.sol
```diff
@@ -209,19 +195,15 @@ contract AzoriusV1 is IAzoriusV1, GuardableModule, Version, UUPSUpgradeable {
...
-        proposals[totalProposalCount].strategy = _strategy;
+        proposals[totalProposalCount].strategy = address(strategy);
         proposals[totalProposalCount].txHashes = txHashes;
         proposals[totalProposalCount].timelockPeriod = timelockPeriod;
         proposals[totalProposalCount].executionPeriod = executionPeriod;
 
-        // not all strategy contracts will necessarily use the txHashes and _data values
-        // they are encoded to support any strategy contracts that may need them
-        IBaseStrategyV1(_strategy).initializeProposal(
-            abi.encode(totalProposalCount, txHashes, _data)
```
adamgall, line 220: Same as previous comment, might want to keep `txHashes` and `_data` for future forward-compatibility

contracts/deployables/modules/AzoriusV1.sol
```diff
@@ -337,63 +277,28 @@ contract AzoriusV1 is IAzoriusV1, GuardableModule, Version, UUPSUpgradeable {
...
         uint32 _proposalId
     ) public view returns (ProposalState) {
+        if (_proposalId >= totalProposalCount) revert InvalidProposal();
         Proposal memory _proposal = proposals[_proposalId];
 
-        if (_proposal.strategy == address(0)) revert InvalidProposal();
-
-        IBaseStrategyV1 _strategy = IBaseStrategyV1(_proposal.strategy);
-        (, uint48 votingEndTimestamp) = _strategy.getVotingTimestamps(
+        (, uint48 votingEndTimestamp) = strategy.getVotingTimestamps(
```
adamgall: Bug: need to check the _proposal's_ strategy, because the contract's strategy might have been replaced while a proposal is active.

adamgall: fixed

contracts/interfaces/decent/deployables/IAzoriusV1.sol
adamgall: fixed

contracts/deployables/modules/AzoriusV1.sol
adamgall: fixed

contracts/interfaces/decent/deployables/IAzoriusV1.sol
```diff
@@ -3,60 +3,22 @@ pragma solidity ^0.8.30;
...
- * Azorius manages the state of Proposals submitted to a DAO, along with the associated strategies
- * ([BaseStrategy](../BaseStrategy.md)) for voting that are enabled on the DAO.
- *
- * Any given DAO can support multiple voting BaseStrategies, and these strategies are intended to be
- * as customizable as possible.
- *
- * Proposals begin in the `ACTIVE` state and will ultimately end in either
- * the `EXECUTED`, `EXPIRED`, or `FAILED` state.
- *
- * `ACTIVE` - a new proposal begins in this state, and stays in this state
```
twodam, line 19: why remove these comments/docs? it seem useful for understanding contracts, something like `as well as parent DAOs time to initiate a freeze`. some of them maybe not necessary, some of them can explain high level interactions.

contracts/interfaces/decent/deployables/IAzoriusV1.sol
```diff
@@ -135,52 +48,10 @@ interface IAzoriusV1 {
...
-     * requires the address of the first strategy you would like, along
-     * with the total count of strategies to return, rather than
-     * returning the whole list at once.
-     *
-     * @param _startAddress contract address of the BaseStrategy to start with
-     * @param _count maximum number of BaseStrategies that should be returned
-     * @return _strategies array of BaseStrategies
-     * @return _next next BaseStrategy contract address in the linked list
-     */
-    function getStrategies(
```
twodam, line 158: maybe add getter of `IStrategyBaseV1 public strategy;`, for example: `function strategy() external view returns (address);`

contracts/deployables/modules/AzoriusV1.sol
```diff
@@ -488,43 +341,21 @@ contract AzoriusV1 is IAzoriusV1, GuardableModule, Version, UUPSUpgradeable {
...
-     * @param _executionPeriod new execution period
-     */
     function _updateExecutionPeriod(uint32 _executionPeriod) internal {
         executionPeriod = _executionPeriod;
         emit ExecutionPeriodUpdated(_executionPeriod);
     }
 
-    /// Implementation for the version
+    function _updateStrategy(address _strategy) internal {
+        strategy = IStrategyBaseV1(_strategy);
```
twodam, line 357: add zero address check for strategy? like we used to check for `enableStrategy` as below code:
```solidity
function enableStrategy(address _strategy) public override onlyOwner {
        if (_strategy == address(0) || _strategy == SENTINEL_STRATEGY)
            revert InvalidStrategy();
```

twodam (changes requested):

twodam:
I haven’t reviewed https://github.com/decentdao/decent-contracts/pull/270 yet, but I’d love to hear more about why a single strategy with multiple token adapters is preferable to using multiple strategies. Is there a blocker with the multiple strategy approach that doesn’t meet our future needs? This could help me better understand the reasoning behind the transition.

contracts/interfaces/decent/deployables/IAzoriusV1.sol
adamgall: You'll see me removing comments throughout all of these PRs. My intention is to do one final PR near then end in which I re-add comments back into ALL of the contracts and interfaces, in order to make sure they're all consistent, formatted the same, use the same kind of language, etc.

contracts/interfaces/decent/deployables/IAzoriusV1.sol
adamgall: That's fair. I'm sure you're aware that the `IStrategyBaseV1 public strategy` variable in `Azorius` automatically exposes a `function strategy() external view returns (address)` function, but you're right that it might be good to explicitly include that in the interface here.

contracts/deployables/modules/AzoriusV1.sol
adamgall: Good call, will do.

contracts/interfaces/decent/deployables/IAzoriusV1.sol
adamgall: You know what, in order to do this "properly", I should include getter functions in the interface for _all_ of the public variables in Azorious (and all of the other contracts), which would then require me to mark all of those corresponding variables in the contracts as private, and then create public getter functions which expose those private variables.

I'm going to put this on my TODO list and tackle it (for all contracts, not just Azorius) in a future PR.

adamgall:
> I haven’t reviewed #270 yet, but I’d love to hear more about why a single strategy with multiple token adapters is preferable to using multiple strategies. Is there a blocker with the multiple strategy approach that doesn’t meet our future needs? This could help me better understand the reasoning behind the transition.

Sure thing.

Originally, a single strategy was responsible for three things:

1. Determining if a `msg.sender` was _allowed_ to create a proposal targeting the given strategy.
2. Determining if a given proposal is currently active or over.
3. Determining if a given proposal has passed or not.

The intention here is to decouple these responsibilities into independent adapters.

So now:

1. We have independent adapters specifically for determining if a user can submit a proposal.
2. The single strategy is responsible for determining if a proposal is active or not. In practice no one wants different proposals to have different periods.
3. We have independent adapters specifically for determining how much weight a given user has for voting, which indirectly plays into understanding if a proposal has passed or not. The core logic for "is passed" lives in the single strategy.

twodam (approved):

tbwebb22 (approved):

PR #287, opened 2025-05-28
feat(VotesERC20LockableV1): add `maxTotalSupply` to cap mintable token supply

twodam:
### Summary

* Added `MaxTotalSupplyUpdated` event, `ExceedMaxTotalSupply` error, and a `maxTotalSupply` getter to the `IMintableV1` interface.
* Updated the `VotesERC20LockableV1` contract:
    * Introduced a `maxTotalSupply` variable to cap the total supply during minting.
    * Integrated `_maxTotalSupply` into the `initialize` function.
    * Enforced the `maxTotalSupply` limit within the `mint` function.
* Extended the `VotesERC20LockableV1.test.ts` test suite with:
    * Test cases for the new `setMaxTotalSupply` function.
    * Validations for the `mint` function to ensure supply limits are respected.


hardhat.config.ts
```diff
@@ -15,6 +15,7 @@ const config: HardhatUserConfig = {
       {
         version: '0.8.30',
         settings: {
+          viaIR: true,
```
adamgall: I don't want this line added. Let's figure out how to get the contracts to compile without it please.

adamgall (changes requested):

hardhat.config.ts
adamgall: I've fixed this

contracts/deployables/erc20/VotesERC20LockableV1.sol
```diff
@@ -77,7 +73,19 @@ contract VotesERC20LockableV1 is ILockableV1, IMintableV1, VotesERC20V1 {
...
+        }
+    }
+
     function mint(address to, uint256 amount) external onlyOwner {
+        uint256 newTotalSupply = totalSupply() + amount;
+        if (newTotalSupply > maxTotalSupply) {
+            revert ExceedMaxTotalSupply();
+        }
         _mint(to, amount);
     }
```
adamgall: Just to be clear, seems like this code allows for setting a new `maxTotalSupply` which is _less than_ the current amount of minted tokens. It would not allow any MORE tokens to be minted, but it does create a strange state.

Is this what we want?

I'm not even sure how we would be able to check the current amount of minted tokens (in order to provide a safety check)...

adamgall: Wait duh, there's a `totalSupply()` function. We could use that to make sure the `newMaxTotalSupply` doesn't exceed current `totalSupply()`.

twodam: yeah, it didn't check if new `maxTotalSupply` is smaller, I thought it may make sense with future burnable feature so that `totalSupply` can decrease. also we can always call another `setMaxTotalSupply` to adjust it, nothing stuck.

twodam: though you are right, this create a strange state, will update.

contracts/deployables/erc20/VotesERC20LockableV1.sol
```diff
@@ -77,10 +77,27 @@ contract VotesERC20LockableV1 is IVotesERC20LockableV1, VotesERC20V1 {
...
+        uint256 newMaxTotalSupply
+    ) external virtual override onlyOwner {
+        if (newMaxTotalSupply < totalSupply()) {
+            revert InvalidMaxTotalSupply();
+        }
+        uint256 currentlyMaxTotalSupply = _maxTotalSupply;
+        _maxTotalSupply = newMaxTotalSupply;
+        if (currentlyMaxTotalSupply != newMaxTotalSupply) {
+            emit MaxTotalSupplyUpdated(newMaxTotalSupply);
+        }
```
adamgall: Nitpick, but why only emit the event if the new value is different than the current value? If you're doing that, I would suggest to include the assignment inside the condition too.

But my preference would be to just remove the condition entirely. Always do the assignment and always emit the event.

adamgall (changes requested):

contracts/deployables/erc20/VotesERC20LockableV1.sol
twodam: I was trying to match the style used in the `whitelist` function above:
```solidity
function whitelist(address account, bool isWhitelisted) external onlyOwner {
        bool currentlyWhitelisted = whitelisted[account];
        whitelisted[account] = isWhitelisted;
        if (currentlyWhitelisted != isWhitelisted) {
            emit Whitelisted(account, isWhitelisted);
        }
    }
```

adamgall: Ah I see, that makes sense. In retrospect I think I'd like both of those snippets to be updated to the style in my first comment then 😅

twodam: I’ve updated `setMaxTotalSupply`. Maybe you can update `whitelist` separately outside this PR? I believe you have some tests where you prefer that event not to be emitted:
```solidity
describe('When adding the same address again', () => {
  beforeEach(async () => {
    addTx = await proxy.connect(owner).whitelist(whitelistMember.address, true);
  });

  it('should not emit an event', async () => {
    await expect(addTx).to.not.emit(proxy, 'Whitelisted');
  });
});
```

adamgall: Will do!

adamgall (approved):

twodam:
@adamgall could you merge this PR? I don't have the access to do it.

adamgall:
I'll do it tomorrow morning (my time), I'll also try to diagnose why you can't merge...

tbwebb22 (approved):
LGTM

PR #296, opened 2025-05-29
`ENG-935` make ERC20 burnable by owner

twodam:
Depends on https://github.com/decentdao/decent-contracts/pull/287

### Summary

- Added a `burn` function that can only be called by the owner.
- Updated the `isTransferable` modifier to allow token burning regardless of whether the token is locked.
- Added some tests related with new `burn` function.


contracts/deployables/erc20/VotesERC20LockableV1.sol
jigglyjams, line 1: We should modify this so an account can only burn their own tokens. The owner of the contract should not be able to burn any accounts tokens.

jigglyjams (changes requested):

contracts/deployables/erc20/VotesERC20LockableV1.sol
```diff
@@ -101,11 +102,15 @@ contract VotesERC20LockableV1 is IVotesERC20LockableV1, VotesERC20V1 {
         _mint(to, amount);
     }
 
+    function burn(address account, uint256 amount) external onlyOwner {
+        _burn(account, amount);
+    }
```
adamgall: As @jigglyjams said, let's remove `onlyOwner`, remove the `account` param, and call `_burn(msg.sender, amount)`

contracts/deployables/erc20/VotesERC20LockableV1.sol
```diff
@@ -18,13 +18,14 @@ contract VotesERC20LockableV1 is IVotesERC20LockableV1, VotesERC20V1 {
...
-    modifier isTransferable(address from) {
+    modifier isTransferable(address from, address to) {
         if (
             _locked &&
             // overrides while locked
             !(from == owner() || // owner can always transfer
                 _whitelisted[from] || // whitelisted addresses can always transfer
-                from == address(0)) // can always mint when locked
+                from == address(0)) && // can always mint when locked
+            to != address(0) // can always burn when locked
```
adamgall, line 28: With the previous suggestion, I think this condition will need to be looked at again...

We probably want to only allow whitelisted accounts to burn when locked.

adamgall (changes requested):

contracts/deployables/erc20/VotesERC20LockableV1.sol
twodam: there's no burn whitelist described in [next ticket](https://linear.app/decent-labs/issue/ENG-936/allow-mint-through-trusted-separate-mint-whitelist-which-the-owner) now, only mint whitelist, so I assume we want to [allow any token holder to burn](https://linear.app/decent-labs/issue/ENG-935/make-erc20-burnable-by-token-holder#comment-0357d0d7)?

adamgall: That's my understanding, yes.

contracts/deployables/erc20/VotesERC20LockableV1.sol
twodam: updated.

contracts/deployables/erc20/VotesERC20LockableV1.sol
twodam: updated.

tbwebb22 (approved):

adamgall (approved):

PR #309, opened 2025-05-30
Feature/eng 987 unstake

tbwebb22:
Addresses [ENG-987](https://linear.app/decent-labs/issue/ENG-987/add-unstake-function)

This PR adds an unstake function that:

- reverts if amount is zero
- reverts if the user's most recent has stake has not elapsed the `minimumStakingPeriod`
- reverts if the unstake amount is greater than the user's staked amount - there is no custom error for this as this will revert from an integer underflow
- burns the user's `VotesERC20Staked` token amount and transfers to them the staked token amount


adamgall (approved):
Approving, but please add a test for unstaking less than full amount.

adamgall (approved):

PR #311, opened 2025-06-02
refactor: Simplify lock and whitelist logic in VotesERC20LockableV1

adamgall:
Depends on https://github.com/decentdao/decent-contracts/pull/310

Fixes [ENG-1060](https://linear.app/decent-labs/issue/ENG-1060/simplify-lock-and-whitelist-functions-in-voteserc20lockablev1)

This commit simplifies the `lock(bool)` and `whitelist(address, bool)` functions in the `VotesERC20LockableV1` contract.

The key changes are:
-   Removed the check in the `lock` function that prevented setting the lock status to its current value. The `CannotSwitchLockState` error has also been removed from the contract and its interface (`IVotesERC20LockableV1`). Setting the lock to its existing state will now succeed silently and still emit the `Locked` event.
-   Removed the conditional check in the `whitelist` function that only emitted the `Whitelisted` event if the whitelist status was actually changed. The `Whitelisted` event will now be emitted every time the `whitelist` function is called, regardless of whether the state changed.

Corresponding test cases in `VotesERC20LockableV1.test.ts` have been updated to reflect these changes:
-   Tests for attempting to set the lock to its current state now expect success instead of a revert.
-   Tests for whitelisting/unwhitelisting an account that is already in that state now expect the `Whitelisted` event to be emitted.

This simplification makes the functions more idempotent in terms of state change (though events are still emitted) and can slightly reduce gas costs by removing conditional checks.

### DIFF
--- a/contracts/deployables/erc20/VotesERC20LockableV1.sol
+++ b/contracts/deployables/erc20/VotesERC20LockableV1.sol
@@ -60,9 +60,6 @@ contract VotesERC20LockableV1 is IVotesERC20LockableV1, VotesERC20V1 {
     }
 
     function lock(bool locked_) external virtual override onlyOwner {
-        if (locked_ == _locked) {
-            revert CannotSwitchLockState(locked_);
-        }
         _locked = locked_;
         emit Locked(_locked);
     }
@@ -71,11 +68,8 @@ contract VotesERC20LockableV1 is IVotesERC20LockableV1, VotesERC20V1 {
         address account,
         bool isWhitelisted
     ) external virtual override onlyOwner {
-        bool currentlyWhitelisted = _whitelisted[account];
         _whitelisted[account] = isWhitelisted;
-        if (currentlyWhitelisted != isWhitelisted) {
-            emit Whitelisted(account, isWhitelisted);
-        }
+        emit Whitelisted(account, isWhitelisted);
     }
 
     function setMaxTotalSupply(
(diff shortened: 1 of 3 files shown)


contracts/deployables/erc20/VotesERC20LockableV1.sol
```diff
@@ -59,9 +59,6 @@ contract VotesERC20LockableV1 is IVotesERC20LockableV1, VotesERC20V1 {
     }
 
     function lock(bool locked_) external virtual override onlyOwner {
-        if (locked_ == _locked) {
-            revert CannotSwitchLockState(locked_);
-        }
```
tbwebb22, line 65: this is another I would consider leaving in - since we know for sure it was a mistake if they pass in a new lock state that's the same as the current lock state

contracts/deployables/erc20/VotesERC20LockableV1.sol
```diff
@@ -70,11 +67,8 @@ contract VotesERC20LockableV1 is IVotesERC20LockableV1, VotesERC20V1 {
         address account,
         bool isWhitelisted
     ) external virtual override onlyOwner {
-        bool currentlyWhitelisted = _whitelisted[account];
         _whitelisted[account] = isWhitelisted;
-        if (currentlyWhitelisted != isWhitelisted) {
-            emit Whitelisted(account, isWhitelisted);
-        }
```
tbwebb22, line 78: same comment here as above

contracts/deployables/erc20/VotesERC20LockableV1.sol
adamgall: yeah, but to argue against that, we don't do any other checks like this when doing setters for things like updating uints.

adamgall: i'd prefer to leave this as is

tbwebb22 (approved):

PR #313, opened 2025-06-02
Feature/eng 988 distribute

tbwebb22:
Addresses [ENG-988](https://linear.app/decent-labs/issue/ENG-988/add-distribute-function)

This PR:
- adds a `distributeRewards()` function that distributes all rewards tokens enabled on this contract
- adds a `distributeRewards(address[] tokens)` function that distributes rewards only for the tokens in the array
- note that these distribute functions do not actually send any tokens to users, they only update accounting to make the distributed tokens claimable by stakers
- these functions do not transfer the rewards tokens into the contract either. New rewards must be sent to the contract before the distribute functions are called. The contract uses its balance of these tokens along with other internal accounting to calculate how much of each token to distribute
- these functions will revert if the total amount staked in the contract is zero - otherwise these rewards would go to no one and would be lost
- added support for native asset, which is specified by passing token address of `0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE`


contracts/deployables/erc20/VotesERC20StakedV1.sol
```diff
@@ -97,6 +102,63 @@ contract VotesERC20StakedV1 is
...
+    }
+
+    function distributeRewards(
+        address[] memory _tokens
+    ) external virtual override {
+        if (_totalStaked == 0) revert ZeroStaked();
+
+        for (uint256 i = 0; i < _tokens.length; ) {
+            if (!_rewardsTokenDatas[_tokens[i]].enabled)
+                revert InvalidRewardsToken();
```
adamgall: Include token address in this error

adamgall (changes requested):
In addition to inline comment, add a view function that returns an array of uints that indicates how much of each rewards token will be "distributed" if the "distribute..." functions are called.

contracts/deployables/erc20/VotesERC20StakedV1.sol
tbwebb22: done

adamgall (approved):

PR #314, opened 2025-06-02
refactor: Consolidate ERC165 supportsInterface logic in AA and Strategy contracts

adamgall:
Depends on https://github.com/decentdao/decent-contracts/pull/312

Fixes [ENG-1071](https://linear.app/decent-labs/issue/ENG-1071/consolidate-and-refine-supportsinterface-in-aa-and-strategy-contracts)

This commit refactors the ERC165 `supportsInterface` implementations across `DecentPaymasterV1`, `SmartAccountValidationV1`, `ERC4337VoterSupportV1`, and `StrategyV1`.

The main changes are:
-   `SmartAccountValidationV1.sol`: Removed its `supportsInterface` function and direct inheritance from `ERC165`. `ERC165` is now expected to be inherited by implementing contracts if needed, or its support is checked at a higher level.
-   `ERC4337VoterSupportV1.sol`: Removed its `supportsInterface` function. Interface support is now handled by `StrategyV1`.
-   `DecentPaymasterV1.sol`: Its `supportsInterface` function now directly checks for `IDecentPaymasterV1`, `ISmartAccountValidationV1`, and `IPaymaster`, then calls `super.supportsInterface()`. The explicit override list `(SmartAccountValidationV1, Version)` has been removed as `Version` (which inherits `ERC165`) is implicitly handled by `super` and `SmartAccountValidationV1`'s interface check is now direct.
-   `StrategyV1.sol`: Its `supportsInterface` function now directly checks for `IStrategyV1`, `IStrategyBaseV1`, `IERC4337VoterSupportV1`, and `ISmartAccountValidationV1`, then calls `super.supportsInterface()`. The explicit override list `(Version, ERC4337VoterSupportV1)` has been removed.

Test updates:
-   `SmartAccountValidationV1.test.ts`: Removed ERC165 tests as the contract no longer implements `supportsInterface` directly.
-   `ERC4337VoterSupportV1.test.ts`: Removed ERC165 tests for the same reason.
-   `StrategyV1.test.ts`: Added tests to confirm `StrategyV1` now reports support for `IERC4337VoterSupportV1` and `ISmartAccountValidationV1`.

This refactoring centralizes the `supportsInterface` logic in the more concrete, top-level contracts, simplifying the intermediate abstract contracts and ensuring that the final deployed contracts accurately report all interfaces they implement through their inheritance chain.

PR #317, opened 2025-06-03
`[ENG-936]` introduce AccessControl to ERC20Lockable with transfer and minter roles

twodam:
Depends on https://github.com/decentdao/decent-contracts/pull/296

### Summary

* Replaced the previous "Transfer Whitelist" mechanism with [AccessControlUpgradeable](https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/master/contracts/access/AccessControlUpgradeable.sol).
* Simplified `isTransferable` modifier by using `hasRole` and grant transfer role to `address(0)`.
* Introduced a `MINTER_ROLE` to manage the minter whitelist.
* The `owner` is granted the [DEFAULT_ADMIN_ROLE](https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/3eee293579c9d248235acadbdc246a46b1cadeab/contracts/access/AccessControlUpgradeable.sol#L46), acting as the default admin for all roles, including transfer and minter roles.


contracts/deployables/erc20/VotesERC20V1.sol
```diff
@@ -100,12 +104,19 @@ contract VotesERC20V1 is
...
+        virtual
+        override(Version, AccessControlUpgradeable)
+        returns (bool)
+    {
         return
             interfaceId == type(IVotesERC20V1).interfaceId ||
             interfaceId == type(IERC20).interfaceId ||
             interfaceId == type(IERC20Permit).interfaceId ||
             interfaceId == type(IVotes).interfaceId ||
+            interfaceId == type(IVersion).interfaceId ||
```
tbwebb22: since this contract inherits from Version, we shouldn't need this line right? I think super.supportsInterface(interfaceId) should handle that for us

contracts/deployables/erc20/VotesERC20V1.sol
```diff
@@ -14,14 +15,16 @@ import {ERC20PermitUpgradeable} from "@openzeppelin/contracts-upgradeable/token/
...
 
 contract VotesERC20V1 is
     IVotesERC20V1,
     Version,
     ERC20VotesUpgradeable,
     ERC20PermitUpgradeable,
     UUPSUpgradeable,
-    Ownable2StepUpgradeable
+    Ownable2StepUpgradeable,
+    AccessControlUpgradeable
```
tbwebb22: Do we need AccessControl inherited in this contract? Doesn't look like we're using it here, and could just inherit it in VotesERC20LockableV1.sol

tbwebb22 (changes requested):

contracts/deployables/erc20/VotesERC20LockableV1.sol
```diff
@@ -22,9 +24,7 @@ contract VotesERC20LockableV1 is IVotesERC20LockableV1, VotesERC20V1 {
         if (
             _locked &&
             // overrides while locked
-            !(from == owner() || // owner can always transfer
-                _whitelisted[from] || // whitelisted addresses can always transfer
-                from == address(0)) && // can always mint when locked
+            !hasRole(TRANSFER_ROLE, from) && // whitelisted addresses can always transfer
             to != address(0) // can always burn when locked
         ) {
             revert IsLocked();
```
tbwebb22: I'm a little confused on how this is working with the minter role. 

Since _mint calls _update, which we've overriden with this modifier. And in this modifier we aren't checking for the MINTER_ROLE. Maybe I'm missing something, but do you mind explaining that?

contracts/deployables/erc20/VotesERC20V1.sol
twodam: Inheriting AccessControl directly in VotesERC20LockableV1 results in a [Linearization of inheritance graph impossible](https://docs.soliditylang.org/en/v0.8.30/contracts.html#multiple-inheritance-and-linearization) error. However, inheriting it in VotesERC20V1 compiles without issues. I’m not sure I fully grasp the [diamond problem](https://en.wikipedia.org/wiki/Multiple_inheritance#The_diamond_problem), but this setup works as expected.
image

contracts/deployables/erc20/VotesERC20V1.sol
twodam: Both parent contracts implement the `supportsInterface` method, which requires using `override(Version, AccessControlUpgradeable)`. Since `AccessControlUpgradeable` is listed last in the inheritance list, the call to `super.supportsInterface` actually resolves to `AccessControlUpgradeable.supportsInterface`.
I have updated this line to be more explicit: `Version.supportsInterface(interfaceId) ||`

contracts/deployables/erc20/VotesERC20LockableV1.sol
twodam: The MINTER_ROLE is enforced in the mint function, as shown below:
```solidity
function mint(
    address to,
    uint256 amount
) external virtual override onlyRole(MINTER_ROLE) {
```
To elaborate further: both transfer and mint ultimately call _update, but transfer explicitly forbids either from or to from being address(0). Therefore, the only valid way to mint tokens is through the mint function, which is why the MINTER_ROLE check is implemented there.

contracts/deployables/erc20/VotesERC20V1.sol
twodam: Maybe we could update `Version` to remove ERC165, similar to what @adamgall did in [this comment](https://github.com/decentdao/decent-contracts/pull/314#issuecomment-2937679873)?
though we may need to all places referencing `Version.supportsInterface`.

adamgall: @twodam i agree with your suggestion. i've just created a PR which removes ERC165 inheritance from Version: https://github.com/decentdao/decent-contracts/pull/319.

After that PR is merged, merge develop into your branch here then move the AccessControl inheritance out of VotesERC20 and into VotesERC20Lockable.

twodam: updated!

contracts/deployables/erc20/VotesERC20LockableV1.sol
```diff
@@ -5,15 +5,24 @@ import {IVotesERC20LockableV1} from "../../interfaces/decent/deployables/IVotesE
 import {IVotesERC20V1} from "../../interfaces/decent/deployables/IVotesERC20V1.sol";
 import {VotesERC20V1} from "./VotesERC20V1.sol";
 import {Version} from "../Version.sol";
-import {ERC165} from "@openzeppelin/contracts/utils/introspection/ERC165.sol";
-
-contract VotesERC20LockableV1 is IVotesERC20LockableV1, VotesERC20V1 {
+import {IAccessControl} from "@openzeppelin/contracts/access/IAccessControl.sol";
+import {ERC165Upgradeable} from "@openzeppelin/contracts-upgradeable/utils/introspection/ERC165Upgradeable.sol";
```
adamgall: What's the reason for using ERC165Upgradeable instead of ERC165?

twodam: updated

contracts/deployables/erc20/VotesERC20LockableV1.sol
```diff
@@ -63,14 +72,6 @@ contract VotesERC20LockableV1 is IVotesERC20LockableV1, VotesERC20V1 {
...
-        address account,
-        bool isWhitelisted
-    ) external virtual override onlyOwner {
-        _whitelisted[account] = isWhitelisted;
-        emit Whitelisted(account, isWhitelisted);
-    }
-
     function setMaxTotalSupply(
         uint256 newMaxTotalSupply
     ) external virtual override onlyOwner {
```
adamgall: Do you think it's weird that we're still using `Ownable` modifiers (from the base VotesERC20 contract) in this contract? I think that's a little weird. I think I'd like to see another role defined here for "modifying state" (of which the `_owner` is added to in initializer), then use that role in place of the two `onlyOwner` modifiers in this contract (`lock()` and `setMaxTotalSupply()`).

@tbwebb22 thoughts?

tbwebb22: hmmm so the owner will pretty much always be the DAO. The idea with roles for minting and transferring is that there could be a need to have multiple addresses that can mint or transfer.

I'm struggling to see a use case for having an address other than the DAO be able to `lock()` or `setMaxSupply()`, can you think of any?

I can also understand the argument for using roles for `lock()` and `setMaxSupply()`  just for consistency however.

adamgall: My reasoning is more so for consistency and separation of concerns.

The base `VotesERC20` is `Ownable` ONLY for the `_authorizeUpgrade()` call. There are no other `onlyOwner` functions.

But then in this contract, we introduce the concept of more refined Access Control, but then _also_ use `onlyOwner` for only _some_ of those authorized functions. That just feels weird to me.

I certainly don't disagree that in practice there will very likely not be a use case where someone other than the DAO is in charge of locking or changing max supply, but I think it's still strange to use `onlyOwner` on those.

Why not do `onlyRole(DEFAULT_ADMIN_ROLE)` for those two functions instead of `onlyOwner`? I think that would achieve the same result, and it would also fully remove `onlyOwner` from this contract.

tbwebb22: ok yeah makes sense to me to maintain consistency, so I agree we should add these two roles

adamgall: I've implemented this

PR #318, opened 2025-06-03
`[ENG-983]` add a to whitelist for transfer

twodam:
### Summary

* Added `TRANSFER_TO_ROLE`, which acts as a recipient (`to`) whitelist. This can be used with staking contracts to ensure token holders can always stake, even when tokens are locked.
* Simplified the `isTransferable` modifier by granting `TRANSFER_TO_ROLE` to `address(0)`.


contracts/deployables/erc20/VotesERC20LockableV1.sol
```diff
@@ -14,6 +14,7 @@ contract VotesERC20LockableV1 is IVotesERC20LockableV1, VotesERC20V1 {
     uint256 internal _maxTotalSupply;
 
     bytes32 public constant TRANSFER_ROLE = keccak256("TRANSFER_ROLE");
```
tbwebb22: maybe we should call this "TRANSFER_FROM_ROLE" to be more explicit

tbwebb22 (approved):

contracts/deployables/erc20/VotesERC20LockableV1.sol
twodam: updated!

adamgall (approved):

PR #325, opened 2025-06-05
Feature/eng 989 claim

tbwebb22:
This PR addresses [ENG-989](https://linear.app/decent-labs/issue/ENG-989/add-claim-function), [ENG-990](https://linear.app/decent-labs/issue/ENG-990/add-claimable-function), and [ENG-991](https://linear.app/decent-labs/issue/ENG-991/add-catch-function-to-reject-erc721-transfer)

This PR:
- adds two claimRewards() functions - one that claims rewards for all the existing rewards tokens, and one that lets the caller specify which rewards tokens to claim for
- adds two claimableRewards() view functions that returns the claimable amounts for the specified staker. One returns for all rewards tokens, the other lets the caller specify which rewards tokens
- fixes a bug in the case that the staked token is also a rewards token. This fix is the else if block added in the _distributableRewards function
- Note that ENG-991 (add catch function to reject ERC721 transfer) is satisfied here by simply not implementing IERC721Receiver
- adds a test helper function `executeTxAndCheckBalanceDeltas` that can be passed a tx to execute, and expected deltas of token or native assets. Test will fail if any of these actual deltas don't match the expected deltas


contracts/deployables/erc20/VotesERC20StakedV1.sol
```diff
@@ -133,6 +133,73 @@ contract VotesERC20StakedV1 is
...
+        for (uint256 i = 0; i < _tokens.length; ) {
+            if (!_rewardsTokenDatas[_tokens[i]].enabled)
+                revert InvalidRewardsToken(_tokens[i]);
+
+            _claimRewards(msg.sender, _recipient, _tokens[i]);
+
+            unchecked {
+                ++i;
+            }
+        }
```
adamgall: Would it save a bit of gas to create a local instance of the `_tokens[i]`, instead of pulling from the array twice?

```solidity
        for (uint256 i = 0; i < _tokens.length; ) {
            address token = _tokens[i];
            if (!_rewardsTokenDatas[token].enabled)
                revert InvalidRewardsToken(token);

            _claimRewards(msg.sender, _recipient, token);

            unchecked {
                ++i;
            }
        }
```

contracts/deployables/erc20/VotesERC20StakedV1.sol
```diff
@@ -385,6 +440,64 @@ contract VotesERC20StakedV1 is
...
+        for (uint256 i = 0; i < _tokens.length; ) {
+            if (!_rewardsTokenDatas[_tokens[i]].enabled)
+                revert InvalidRewardsToken(_tokens[i]);
+
+            claimableRewards_[i] = _claimableRewards(_staker, _tokens[i]);
+
+            unchecked {
+                ++i;
+            }
+        }
```
adamgall: Same question here

adamgall (approved):
Other than a couple of questions about potential(?) gas savings, lgtm

contracts/deployables/erc20/VotesERC20StakedV1.sol
tbwebb22: I don't think this would be cheaper, but I'll try to run some gas comparisons to find out

tbwebb22: ok it did actually save a bit of gas, so I committed this change in both places

Before:
claimRewards(address,address[]): 66,321 - 205,696 gas (avg: 134,754)


After:
claimRewards(address,address[]): 66,282 - 205,618 gas (avg: 134,676)

PR #336, opened 2025-06-07
refactor(ProxyFactory): Remove deployer-specific salt logic

adamgall:
Depends on https://github.com/decentdao/decent-contracts/pull/335

Fixes [ENG-1115](https://linear.app/decent-labs/issue/ENG-1115/simplify-proxyfactory-by-removing-deployer-specific-salt-logic)

The `ProxyFactory` has been simplified by removing the logic that namespaces the `salt` with the deployer's address. Previously, the factory would combine `msg.sender` with the provided salt to generate a unique salt for each deployer.

This change makes the factory's behavior more aligned with the standard `CREATE2` opcode, where the resulting contract address is determined only by the factory's address, the salt, and the contract's creation code. The `deployer` is no longer a factor.

Key changes:
- `deployProxy` now uses the provided salt directly.
- `predictProxyAddress` no longer requires a `deployer_` parameter.
- Tests have been updated to reflect this simplified, deployer-agnostic behavior. The test case for multiple deployers creating different addresses with the same salt has been removed, as this is no longer the expected outcome.

### DIFF
--- a/contracts/interfaces/decent/singletons/IProxyFactory.sol
+++ b/contracts/interfaces/decent/singletons/IProxyFactory.sol
@@ -13,7 +13,6 @@ interface IProxyFactory {
     function predictProxyAddress(
         address implementation_,
         bytes calldata initData_,
-        bytes32 salt_,
-        address deployer_
+        bytes32 salt_
     ) external view returns (address proxy);
 }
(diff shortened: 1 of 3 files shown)


adamgall:
@tbwebb22 let's definitely talk about this one

tbwebb22 (approved):

PR #347, opened 2025-06-08
feat(governance): Implement future-dated proposals

adamgall:
Depends on https://github.com/decentdao/decent-contracts/pull/346

Fixes [ENG-1126](https://linear.app/decent-labs/issue/ENG-1126/feature-implement-future-dated-proposals)

Adds support for creating proposals with a start time in the future. This allows for proposals to be submitted now but only become active for voting at a specified later time.

This change introduces a `PENDING` state to the proposal lifecycle and a `proposalStartTime` parameter to the proposal creation process. The system now includes robust checks to prevent voting on pending proposals and to handle state transitions correctly.

To support off-chain validation for Account Abstraction, this feature also introduces state tracking for vote attempts made before the voting period begins (`voteCastedBeforeVotingPeriodStarted`). This allows bundlers to simulate and predict the validity of vote transactions without relying on `block.timestamp`.

- Introduced a `PENDING` state in the `ProposalState` enum.
- Added `proposalStartTime` parameter to `submitProposal` and `initializeProposal` functions.
- Implemented logic in `StrategyV1` and `ModuleAzoriusV1` to handle the new `PENDING` state and future start times.
- Added `voteCastedBeforeVotingPeriodStarted` state and `VotingPeriodNotStarted` event for reliable off-chain validation.
- Updated all relevant tests to cover the full lifecycle of future-dated proposals, including state transitions and voting restrictions.

### DIFF
--- a/contracts/deployables/modules/ModuleAzoriusV1.sol
+++ b/contracts/deployables/modules/ModuleAzoriusV1.sol
@@ -159,11 +159,12 @@ contract ModuleAzoriusV1 is
         Proposal memory _proposal = _proposals[proposalId_];
         IStrategyV1 strategy_ = IStrategyV1(_proposal.strategy);
 
-        (, uint48 votingEndTimestamp) = strategy_.getVotingTimestamps(
-            proposalId_
-        );
+        (uint48 votingStartTimestamp, uint48 votingEndTimestamp) = strategy_
+            .getVotingTimestamps(proposalId_);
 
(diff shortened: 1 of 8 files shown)


adamgall:
Closing this, there is a bug in the gasless voting `validStrategyVote()` function. If a vote was attempted before a proposal is started, then _any_ gasless vote attempt after that moment, including gasless vote attempts that happen during the actual proposal active time, will fail.

PR #351, opened 2025-06-10
Feature/eng 1040 countersign basic implementation

tbwebb22:
This PR creates the foundation for the countersigning contract, addressing [ENG-1040](https://linear.app/decent-labs/issue/ENG-1042/pass-in-actions-to-be-performed) and [ENG-1042](https://linear.app/decent-labs/issue/ENG-1042/pass-in-actions-to-be-performed)

- Currently only includes constructor to setup contract state, and view functions to read that state
- Uses `pragma abicoder v2` to enable writing from memory Transaction struct to storage Transaction struct

### DIFF
--- a/contracts/deployables/countersign/CountersignV1.sol
+++ b/contracts/deployables/countersign/CountersignV1.sol
@@ -4,15 +4,16 @@ pragma solidity ^0.8.30;
 import {IVersion} from "../../interfaces/decent/deployables/IVersion.sol";
 import {ICountersignV1} from "../../interfaces/decent/deployables/ICountersignV1.sol";
 import {ERC165} from "@openzeppelin/contracts/utils/introspection/ERC165.sol";
+import {Initializable} from "@openzeppelin/contracts/proxy/utils/Initializable.sol";
 
-contract CountersignV1 is ICountersignV1, IVersion, ERC165 {
+contract CountersignV1 is ICountersignV1, IVersion, ERC165, Initializable {
     // ======================================================================
     // STATE VARIABLES
     // ======================================================================
 
     string internal _agreementUri;
-    address internal immutable _verificationContract;
-    uint256 internal immutable _minWeight;
+    address internal _verificationContract;
+    uint256 internal _minWeight;
     address[] internal _signerAddresses;
     mapping(address signer => Signer signerData) internal _signerData;
     Transaction[] internal _preExecutionTransactions;
@@ -21,13 +22,17 @@ contract CountersignV1 is ICountersignV1, IVersion, ERC165 {
     // CONSTRUCTOR & INITIALIZERS
     // ======================================================================
 
-    constructor(
+    constructor() {
+        _disableInitializers();
+    }
+
+    function initialize(
         string memory agreementUri_,
         address verificationContract_,
         uint256 minWeight_,
         SignerInitialization[] memory signerInitializations_,
         Transaction[] memory preExecutionTransactions_
-    ) {
+    ) public virtual override initializer {
         _agreementUri = agreementUri_;
         _verificationContract = verificationContract_;
         _minWeight = minWeight_;
--- a/contracts/deployables/countersign/CountersignV1.sol
+++ b/contracts/deployables/countersign/CountersignV1.sol
@@ -25,49 +25,39 @@ contract CountersignV1 is ICountersignV1, IVersion, ERC165 {
         string memory agreementUri_,
         address verificationContract_,
         uint256 minWeight_,
-        address[] memory signerAddresses_,
-        bool[] memory signerRequired_,
-        uint256[] memory signerWeights_,
-        Transaction[][] memory signerTransactions_,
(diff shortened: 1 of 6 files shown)


contracts/deployables/countersign/CountersignV1.sol
```diff
@@ -0,0 +1,162 @@
...
+    // ======================================================================
+
+    constructor(
+        string memory agreementUri_,
+        address verificationContract_,
+        uint256 minWeight_,
+        address[] memory signerAddresses_,
+        bool[] memory signerRequired_,
+        uint256[] memory signerWeights_,
+        Transaction[][] memory signerTransactions_,
```
adamgall: Can you make a struct for these inputs, so that we can consolidate them down into a single array property? Then you could also get rid of the InvalidArrayWeights error.

hardhat.config.ts
```diff
@@ -19,6 +19,7 @@ const config: HardhatUserConfig = {
             enabled: true,
             runs: 200,
           },
+          viaIR: true,
```
adamgall: I don't love that this is being added. Can you write the code to avoid this?

contracts/deployables/countersign/CountersignV1.sol
```diff
@@ -0,0 +1,162 @@
...
+        _signerAddresses = signerAddresses_;
+        _preExecutionTransactions = preExecutionTransactions_;
+
+        for (uint256 i = 0; i < signerAddresses_.length; ) {
+            _signerData[signerAddresses_[i]] = Signer({
+                isSigner: true,
+                required: signerRequired_[i],
+                signed: false,
+                weight: signerWeights_[i],
+                transactions: signerTransactions_[i]
```
adamgall: I think this is the line that was requiring the `viaIR`, right? Maybe this can be broken out into its own loop as well... kinda shitty though.

i guess i don't really know the tradeoffs of `viaIR: true`. Why wouldn't it just be "true" by default all the time? there must be a tradeoff.

contracts/deployables/countersign/CountersignV1.sol
```diff
@@ -0,0 +1,174 @@
...
+    // CONSTRUCTOR & INITIALIZERS
+    // ======================================================================
+
+    constructor(
+        string memory agreementUri_,
+        address verificationContract_,
+        uint256 minWeight_,
+        SignerInitialization[] memory signerInitializations_,
+        Transaction[] memory preExecutionTransactions_
+    ) {
```
adamgall: This constructor should be an `initialize` function so that instances of this contract can be deployed the same was as the other contracts in this repo.

hardhat.config.ts
```diff
@@ -18,7 +18,7 @@ const config: HardhatUserConfig = {
           optimizer: {
             enabled: true,
             runs: 200,
-          },
+          }
```
adamgall: and plz undo the change on this line

adamgall (changes requested):

contracts/deployables/countersign/CountersignV1.sol
tbwebb22: this is fixed

hardhat.config.ts
tbwebb22: done

test/deployables/countersign/Countersign.test.ts
```diff
@@ -9,9 +9,49 @@ import {
...
+        ],
+        [
+          agreementUri,
+          verificationContract,
+          minWeight,
+          signerInitializations,
+          preExecutionTransactions,
+        ],
+      )
+      .slice(2);
```
adamgall, line 46: You can simplify this and get type safety:

```ts
const fullInitData = CountersignV1__factory.createInterface().encodeFunctionData('initialize', [
    agreementUri,
    verificationContract,
    minWeight,
    signerInitializations,
    preExecutionTransactions,
  ]);
```

adamgall (approved):
Approving, but one suggestion

test/deployables/countersign/Countersign.test.ts
adamgall: looks like there are quite a few test files in this repo that do it the dumb way huh

adamgall: here i fixed em: https://github.com/decentdao/decent-contracts/pull/355

contracts/deployables/countersign/CountersignV1.sol
twodam: > I think this is the line that was requiring the `viaIR`, right? Maybe this can be broken out into its own loop as well... kinda shitty though.
> 
> i guess i don't really know the tradeoffs of `viaIR: true`. Why wouldn't it just be "true" by default all the time? there must be a tradeoff.

Here's a ChatGPT deep research on this `viaIR` if you are interested: https://chatgpt.com/s/dr_68538ac594608191b7dc299d6f325b4e

PR #350, opened 2025-06-10
refactor(voting): Rename initialization functions for consistency

adamgall:
This commit updates the initialization function names across various voting-related contracts to ensure uniformity and clarity. The changes include renaming `__BaseFreezeVotingV1_init` to `__FreezeVotingBaseV1_init` in `FreezeVotingAzoriusV1` and `FreezeVotingMultisigV1`, as well as renaming `__BaseVotingAdapterV1_init` to `__VotingAdapterBaseV1_init` in the voting adapter contracts. This refactoring aims to enhance code readability and maintainability.

PR #352, opened 2025-06-10
refactor(module-azorius): Decouple strategy from initialization

adamgall:
Depends on https://github.com/decentdao/decent-contracts/pull/350

Fixes [ENG-1146](https://linear.app/decent-labs/issue/ENG-1146/refactor-moduleazoriusv1-to-decouple-strategy-from-initialization)

This change refactors the `ModuleAzoriusV1` contract to remove the strategy from its `initialize` and `setUp` functions. The strategy is now intended to be set in a subsequent, separate transaction by calling the `updateStrategy` function.

This decoupling is a critical prerequisite for enabling secure, atomic deployments of the entire Azorius system. By breaking the direct initialization dependency, it allows deployment workflows (using tools like `MultiSend` or a dedicated deployer contract) to configure the module in a more linear, two-step process (`initialize`, then `updateStrategy`). This is particularly important for handling the complex ownership hand-offs required when deploying via a `ProxyFactory`, where the initial owner (`ProxyFactory`) must transfer ownership to an orchestrator contract (e.g., `MultiSend`) which then performs the final configuration.

- Removed the `strategy` parameter from the `initialize` and `setUp` functions in `ModuleAzoriusV1.sol`.
- Updated the `IModuleAzoriusV1` interface to reflect the new `initialize` function signature.
- Aligned the `MockModuleAzoriusV1` contract with the updated interface to ensure test compatibility.
- Reworked the tests for `ModuleAzoriusV1` to use the new two-step configuration flow, first initializing the contract and then calling `updateStrategy`.

### DIFF
--- a/contracts/deployables/freeze-voting/FreezeVotingAzoriusV1.sol
+++ b/contracts/deployables/freeze-voting/FreezeVotingAzoriusV1.sol
@@ -42,7 +42,7 @@ contract FreezeVotingAzoriusV1 is
         address parentAzorius_,
         address lightAccountFactory_
     ) public virtual override initializer {
-        __BaseFreezeVotingV1_init(
+        __FreezeVotingBaseV1_init(
             owner_,
             freezeProposalPeriod_,
             freezePeriod_,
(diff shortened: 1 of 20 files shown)


adamgall:
Closing this PR, it was opened prematurely, before I had a good handle on the root of the problem I was trying to solve. Will re-open with a better solution hopefully soon.

PR #362, opened 2025-06-15
feat(core): Introduce SystemDeployer for atomic DAO deployment

adamgall:
Depends on https://github.com/decentdao/decent-contracts/pull/369

Fixes [ENG-1164](https://linear.app/decent-labs/issue/ENG-1164/implement-systemdeployer-for-atomic-one-shot-dao-creation)

feat(core): Introduce SystemDeployerV1 and SystemDeployerEventEmitterV1 for atomic DAO deployment

This commit introduces the `SystemDeployerV1` contract, a powerful new component designed to deploy and configure an entire DAO governance system in a single, atomic transaction.

Previously, setting up a full DAO required a complex and error-prone sequence of manual deployments and configuration calls. The `SystemDeployerV1` solves this by orchestrating the entire process.

The deployment flow is initiated from the `SafeProxyFactory`, which creates a new Safe proxy and calls its `setup` function. The Safe's `setup` then makes a `delegatecall` to the `SystemDeployerV1.setupSafe` function. This grants the `SystemDeployerV1` the context of the new Safe, allowing it to:
- Deploy all necessary governance contracts (Tokens, Strategy, Modules, Adapters, Guards) as CREATE2 proxies with a shared salt for deterministic addresses.
- Correctly initialize and link all components, managing complex inter-dependencies, such as the two-step initialization required for `StrategyV1`.
- Enable the deployed modules on the Safe and set guards.

After successfully deploying the system, the `SystemDeployerV1` calls a new singleton contract, `SystemDeployerEventEmitterV1`, to emit an event from the new Safe (the context of the `SystemDeployerV1`), the `SafeProxyFactory` contract, the `salt` used for deployment, and the `initData` used for initialization.

This new architecture drastically simplifies DAO creation, making it more reliable, gas-efficient, and user-friendly. The commit includes the `SystemDeployerV1.sol` contract, the `SystemDeployerEventEmitterV1.sol` contract, and an extensive test suite (`SystemDeployerV1.test.ts`,`SystemDeployerEventEmitterV1.test.ts`) to ensure the robustness of various deployment configurations.

### DIFF
--- a/contracts/interfaces/decent/singletons/ISystemDeployerV1.sol
+++ b/contracts/interfaces/decent/singletons/ISystemDeployerV1.sol
@@ -1,0 +1,199 @@
+// SPDX-License-Identifier: AGPL-3.0
+pragma solidity ^0.8.30;
+
+import {IVotesERC20V1} from "../deployables/IVotesERC20V1.sol";
+
+interface ISystemDeployerV1 {
+    // --- Errors ---
+
+    error ImplementationMustBeAContract();
+    error VotesERC20TokenNotFoundAtIndex(uint256 typeIndex, uint256 tokenIndex);
+    error CannotDeployBothFreezeVotingContracts();
+    error FreezeVotingContractNotDeployed();
+    error AzoriusModuleNotDeployed();
+
+    // --- Structs ---
+
+    struct TypeTokenIndex {
+        uint256 typeI;
+        uint256 tokenI;
+    }
+
+    struct VotesERC20V1Params {
+        address implementation;
+        IVotesERC20V1.Metadata metadata;
+        IVotesERC20V1.Allocation[] allocations;
+        uint256 safeSupply;
+    }
+
+    struct VotesERC20LockableV1Params {
+        address implementation;
+        IVotesERC20V1.Metadata metadata;
+        IVotesERC20V1.Allocation[] allocations;
+        bool locked;
+        uint256 maxTotalSupply;
+        uint256 safeSupply;
+    }
+
+    struct VotesERC20Params {
+        VotesERC20V1Params[] votesERC20V1Params;
+        VotesERC20LockableV1Params[] votesERC20LockableV1Params;
+    }
+
+    struct ProposerAdapterERC20V1Params {
+        address implementation;
+        address token;
+        TypeTokenIndex index;
+        uint256 proposerThreshold;
+    }
+
+    struct ProposerAdapterERC721V1Params {
+        address implementation;
+        address token;
+        uint256 proposerThreshold;
+    }
+
+    struct ProposerAdapterHatsV1Params {
+        address implementation;
+        address hatsContract;
+        uint256[] whitelistedHatIds;
+    }
+
+    struct ProposerAdapterParams {
+        ProposerAdapterERC20V1Params[] proposerAdapterERC20V1Params;
+        ProposerAdapterERC721V1Params[] proposerAdapterERC721V1Params;
+        ProposerAdapterHatsV1Params[] proposerAdapterHatsV1Params;
+    }
+
+    struct StrategyV1Params {
+        address implementation;
+        uint32 votingPeriod;
+        uint256 quorumThreshold;
+        uint256 basisNumerator;
+        address lightAccountFactory;
+    }
+
+    struct VotingAdapterERC20V1Params {
+        address implementation;
+        address token;
+        TypeTokenIndex index;
+        uint256 weightPerToken;
+    }
+
+    struct VotingAdapterERC721V1Params {
+        address implementation;
+        address token;
+        uint256 weightPerToken;
+    }
+
+    struct VotingAdapterParams {
+        VotingAdapterERC20V1Params[] votingAdapterERC20V1Params;
+        VotingAdapterERC721V1Params[] votingAdapterERC721V1Params;
+    }
+
+    struct ModuleAzoriusV1Params {
+        address implementation;
+        uint32 timelockPeriod;
+        uint32 executionPeriod;
+    }
+
+    struct AzoriusGovernanceParams {
+        ProposerAdapterParams proposerAdapterParams;
+        StrategyV1Params strategyV1Params;
+        VotingAdapterParams votingAdapterParams;
+        ModuleAzoriusV1Params moduleAzoriusV1Params;
+    }
+
+    struct ModuleFractalV1Params {
+        address implementation;
+        address owner;
+    }
+
+    struct FreezeGuardMultisigV1Params {
+        address implementation;
+        address owner;
+        uint32 timelockPeriod;
+        uint32 executionPeriod;
+    }
+
+    struct FreezeGuardAzoriusV1Params {
+        address implementation;
+        address owner;
+    }
+
+    struct FreezeVotingMultisigV1Params {
+        address implementation;
+        address owner;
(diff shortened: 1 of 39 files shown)


twodam:
why the base branch didn't automatically changed to `develop` after `add-erc165-and-iversion-to-keyvaluepairs` branch got merged?

contracts/interfaces/decent/singletons/ISystemDeployerV1.sol
```diff
@@ -0,0 +1,199 @@
...
+    struct FreezeParams {
+        FreezeGuardParams freezeGuardParams;
+        FreezeVotingParams freezeVotingParams;
+    }
+
+    // --- Events ---
+
+    event ProxyDeployed(address indexed proxy, address indexed implementation);
+
+    event SystemDeployed(
```
twodam, line 163: Just a thought—would SafeDeployed be a clearer name? When I first saw System, I wasn’t quite sure what it referred to until I saw it came from the setupSafe function. Could help make things more immediately clear for others too. Or alternatively, maybe setupSafe could be renamed to setupSystem for consistency?

contracts/interfaces/decent/singletons/ISystemDeployerV1.sol
```diff
@@ -0,0 +1,199 @@
...
+        IVotesERC20V1.Allocation[] allocations;
+        bool locked;
+        uint256 maxTotalSupply;
+        uint256 safeSupply;
+    }
+
+    struct VotesERC20Params {
+        VotesERC20V1Params[] votesERC20V1Params;
+        VotesERC20LockableV1Params[] votesERC20LockableV1Params;
+    }
```
twodam, line 41: Curious—why are we using an array of params here? Is it for deploying multiple ERC20 tokens at once, or more for future flexibility? Also, since SystemDeployer currently only handles ERC20s, do you think it makes sense to start considering ERC721 support too, especially with the frontend already supporting NFT-based governance?

contracts/singletons/SystemDeployerV1.sol
```diff
@@ -0,0 +1,883 @@
...
+            allNewVotesERC20Addresses
+        );
+
+        address azoriusModuleAddress = _deployAzoriusGovernance(
+            salt_,
+            azoriusGovernanceParams_,
+            allNewVotesERC20Addresses
+        );
+
+        _deployModuleFractal(salt_, moduleFractalV1Params_);
```
twodam, line 128: I know we used to go by Fractal, and the contracts still reflect that. With this new PR, do you think it might be a good time to start considering a rename to Decent?

twodam (approved):
This is a relatively big PR—hope it's not too annoying that I'm leaving some comments on naming and a few suggestions that might not be strictly within the current scope.

For the core implementation, especially where multiple modules are being wired together, I found it a bit tricky to follow. Personally, I’m familiar with some of these modules and what they do, but the details tend to fade over time. Having a diagram or a short explanation of how these modules interact could really help—both for people who aren’t closely following day-to-day and for future audits or onboarding.

Also, just want to say this PR is a solid improvement—it’s a great idea to encapsulate the system deployment logic in a dedicated deployer contract, rather than relying on frontend devs to manually construct everything.

contracts/interfaces/decent/singletons/ISystemDeployerV1.sol
adamgall: It's good to question these naming decisions.

The `setupSafe()` function is executing via a delegatecall from the Safe itself, so `address(this)` from within the context of that function _is_ the new Safe address. I could see the argument for renaming this to `setupSystem()`. Or maybe even `setupDecent()` lol.

As for the event name, maybe rather than saying `Deployed` we could say `Setup` (or is it `SetUp`)? So for example, `event SystemSetUp`, or `event DecentSetUp`.

This would probably make it clear that the system has completed its set up functionality.

contracts/interfaces/decent/singletons/ISystemDeployerV1.sol
adamgall: These arrays are exactly for the ability to deploy multiple VotesERC20 and/or VotesERC20Lockable tokens at once. Also notice how this `VotesERC20Params` struct lives _outside_ of the `AzoriusGovernanceParams` struct at the top level of the `setupSafe` function, meaning that this allows for someone to deploy tokens on a new multisig (non-azorius) Safe!

The test file is pretty comprehensive of the different behaviors that are allowed when using this.

But a good rule of thumb to understand the flexibility here is that all 4 of the top level struct params in `setupSafe` are independent and optional.

contracts/singletons/SystemDeployerV1.sol
adamgall: There is _one_ contract still with the word `Fractal` in it: The `ModuleFractalV1` contract. I don't plan on renaming that, it is the module that enables a parent to have control over a child, and i think the name encompasses the purpose quite nicely.

adamgall:
> For the core implementation, especially where multiple modules are being wired together, I found it a bit tricky to follow. Personally, I’m familiar with some of these modules and what they do, but the details tend to fade over time. Having a diagram or a short explanation of how these modules interact could really help—both for people who aren’t closely following day-to-day and for future audits or onboarding.

I agree some documentation would be helpful here. I'll try to get some docs added into the contract when/if I have time later today.

But quickly:

`votesERC20Params_`: two arrays of initialization data for new `VotesERC20V1` and/or new `VotesERC20LockableV1` tokens to deploy.

`moduleFractalV1Params_`: initialization data needed to deploy a `ModuleFractalV1` contract then attach it to the Safe (`address(this)`) as a module.

`freezeParams_`: initialization data for deploying FreezeGuard and FreezeVoting contracts. You can install zero, one, or both FreezeGuard contracts (`FreezeGuardMultisigV1` will be installed as a guard on the Safe, `FreezeGuardAzoriusV1` will be installed on the new `ModuleAzoriusV1` module, assuming it has also been created (and throws an error if not)). Each FreezeGuard requires one and only one FreezeVoting contract.

`azoriusGovernanceParams_`: initialization data for a `ModuleAzoriusV1` contract, a `StrategyV1` contract, one or more ProposerAdapters, and one or more VotingAdapters. The `ProposerAdapterERC20V1` and `VotingAdapterERC20V1` contracts specifically require an ERC20 token, and the configuration data for those contracts allows the caller to pass in an _existing_ ERC20 token address, OR an index to one of the new `VotesERC20` tokens from the `votesERC20Params` list.

contracts/singletons/SystemDeployerV1.sol
twodam: make sense!

contracts/singletons/SystemDeployerEventEmitterV1.sol
```diff
@@ -0,0 +1,51 @@
...
+
+    // --- State-Changing Functions ---
+
+    function emitSystemDeployed(
+        address safeProxyFactory_,
+        bytes32 salt_,
+        bytes calldata initData_
+    ) public virtual override {
+        emit SystemDeployed(msg.sender, safeProxyFactory_, salt_, initData_);
+    }
```
tbwebb22, line 25: One check we could do here to make it harder for someone to manipulate this, is call `deploymentBlock()` on msg.sender (the new Safe), make sure its equal to current block.

tbwebb22 (approved):

PR #371, opened 2025-06-17
`[ENG-1176]` Token should be unlockable from locked state, but not lockable from unlocked state

twodam:
### Context

It doesn't make sense to lock token after it's unlocked.

image


adamgall (commented):
The linked Linear ticket doesn't mention anything about defaulting the token to be locked initially, but it looks like that's what this PR does (in addition to not allow it to be locked after it's been unlocked). Considering that there is additional functionality in this contract (minting, burning, updating total supply), is there a use case of the user wanting to utilize the additional functionality from this contract, but also have it be unlocked from the start? If so, this PR doesn't support that use case.

johnqh:
Lots of changes on requirements. (Even new ones from last night conversation with Parker and lawyer) We will have to use a different initial settings. I am still waiting for things to settle down. Hold on to this PR for the time being. I will update everyone

johnqh (approved):

PR #390, opened 2025-06-20
move non-deterministic initializer args into initialize2

tbwebb22:
This PR updates `VotesERC20Staked` initialize arguments to make its deployment deterministic based on Safe address (owner) and staked token address. 

Name and symbol metadata are removed from initialize, and "Staked " and "st" are prefixed from the staked tokens name and symbol.

`minimumStakingPeriod` and `rewardsTokens` are moved into a new `initialize2` function


adamgall (approved):
Code looks fine.

Can you talk through how this is expected to be deployed in a production environment safely and without initialze2 being frontran?

tbwebb22:
> Code looks fine.
> 
> Can you talk through how this is expected to be deployed in a production environment safely and without initialze2 being frontran?

yeah great question. So we definitely want to deploy the proxy, call initialize, and call initialize2 all atomically. So DAO proposals that deploy this staking contract will also need to include a transaction for calling the initialize2 function so it all happens atomically.

Here's an idea: we could add a variation of `deployProxy` to the `SystemDeployer` that is specifically for contracts that have `initialize2` functions. The Safe would then do a single delegatecall to this function to fully deploy and initialize the staking contract

adamgall:
I like that idea. Or actually just add another param to the current deployProxy function that takes an array of bytes then we loop through them. Do that in a separate PR though.

adamgall:
> I like that idea. Or actually just add another param to the current deployProxy function that takes an array of bytes then we loop through them. Do that in a separate PR though.

Actually i take this all back. I just implemented it and I don't love it. Since that "extra init data" doesn't have anything to do with the deterministic proxy addresses, let's just put that concern to make sure to call "initialize2" onto the client.

PR #411, opened 2025-06-26
Feature/warrant

tbwebb22:
This PR adds abstract contract WarrantBase.sol and WarrantHedgeyV1.sol

WarrantBase
- Includes generic warrant logic supporting both relative and absolute time
- In relative time mode, the vesting token must be a VotesERC20V1, and vesting start will be relative to the token unlock timestamp. Warrant can only be executed after the token has been unlocked.
- In absolute mode, a timestamp is passed in upon initialization for when vesting can begin
- Upon execution a fee is charged to the caller calculated by `tokenAmount` * `tokenPrice`
- Warrants have an expiration that can also be absolute or relative to token unlock time. After expiration, the warrant cannot be executed
- Clawback can be called by the owner once the warrant has expired. This sends the tokens held in the warrant to a recipient address specified by the owner

WarrantHedgeyV1
- Extends WarrantBase to integrate with Hedgey for token lockups
- Includes a relative cliff time that is used with the start time to calculate the absolute cliff time passed into Hedgey
- Includes Hedgey param validation to revert on zero amounts and if calculated cliff time exceeds the calculated vesting end time
- Upon execution, tokens are approved for transfer to Hedgey, and `createPlan` is called on the Hedgey contract which transfers the tokens from this contract to Hedgey, and updates state with the new lockup plan


contracts/deployables/warrant/WarrantBase.sol
```diff
@@ -0,0 +1,315 @@
...
+     * @dev Returns the storage struct for WarrantBase
+     * Following the EIP-7201 namespaced storage pattern to avoid storage collisions
+     */
+    function _getWarrantBaseStorage()
+        internal
+        pure
+        returns (WarrantBaseStorage storage)
+    {
+        WarrantBaseStorage storage $;
+        assembly {
```
adamgall, line 90: Please add `// solhint-disable-next-line no-inline-assembly` above this line

contracts/deployables/warrant/WarrantBase.sol
```diff
@@ -0,0 +1,315 @@
...
+    /**
+     * @dev Returns the storage struct for WarrantBase
+     * Following the EIP-7201 namespaced storage pattern to avoid storage collisions
+     */
+    function _getWarrantBaseStorage()
+        internal
+        pure
+        returns (WarrantBaseStorage storage)
+    {
+        WarrantBaseStorage storage $;
```
adamgall: I know the pattern in this codebase is to never have named return variables in functions in implementation contracts, but these `_getxxxStorage()` functions are the exception to that rule.

Please move the `$` up into the returns line and remove `WarrantBaseStorage storage $;`

Then you can remove the explicit `return $;` as well.

Thanks!

contracts/deployables/warrant/WarrantBase.sol
```diff
@@ -0,0 +1,315 @@
...
+     * @param owner_ Owner address who can clawback after expiration
+     * @param warrantHolder_ Address authorized to execute the warrant
+     * @param token_ Token to be vested
+     * @param feeToken_ Token used for fee payment
+     * @param tokenAmount_ Amount of tokens to vest
+     * @param tokenPrice_ Price per token in fee token units (18 decimals)
+     * @param feeReceiver_ Address that receives fee payments
+     * @param expiration_ Expiration timestamp or duration
+     */
+    function __WarrantBase_init(
```
adamgall, line 118: Please add `// solhint-disable-previous-line func-name-mixedcase` below this function declaration (NOT above it, as that will break the natspec)

contracts/deployables/warrant/WarrantBase.sol
```diff
@@ -0,0 +1,315 @@
...
+                ERC165(token_).supportsInterface(
+                    type(IVotesERC20V1).interfaceId
+                )
+            returns (bool result) {
+                supported = result;
+            } catch {
+                // supported is already false by default
+            }
+            if (!supported) revert UnsupportedToken();
+        }
```
adamgall: What are your thoughts on just removing this? AFAIK the patterns in the rest of the codebase don't really bother with any checks on input addresses like this.

contracts/deployables/warrant/WarrantBase.sol
```diff
@@ -0,0 +1,315 @@
...
+
+        // Calculate and collect fee
+        uint256 feeAmount = ($.tokenAmount * $.tokenPrice) / PRECISION;
+        IERC20($.feeToken).safeTransferFrom(
+            msg.sender,
+            $.feeReceiver,
+            feeAmount
+        );
+
+        // Mark as executed before external call
```
adamgall: comment nitpick: the call to `_executeWarrant()` isn't an external call

contracts/deployables/warrant/WarrantBase.sol
```diff
@@ -0,0 +1,315 @@
...
+            if (
+                block.timestamp >
+                IVotesERC20V1($.token).getUnlockTime() + $.expiration
+            ) revert Expired();
+        } else {
+            if (block.timestamp > $.expiration) revert Expired();
+        }
+
+        // Calculate and collect fee
+        uint256 feeAmount = ($.tokenAmount * $.tokenPrice) / PRECISION;
```
adamgall: It's not clear to me how this equation results in a `feeAmount`... what is this?

Is the variable name `$.tokenPrice` descriptive enough? Maybe it should be called `$.tokenPricePerFeeToken`, or something?

contracts/deployables/warrant/WarrantBase.sol
```diff
@@ -0,0 +1,315 @@
...
+    /**
+     * @inheritdoc IWarrantBase
+     * @dev Validates conditions, collects fee, and delegates to _executeWarrant
+     */
+    function execute(address recipient_) public virtual override {
+        WarrantBaseStorage storage $ = _getWarrantBaseStorage();
+
+        // Validate caller and state
+        if (msg.sender != $.warrantHolder) revert OnlyWarrantHolder();
+        if (recipient_ == address(0)) revert AddressZero();
```
adamgall: Maybe remove this check? Or, add it into `clawback(address recipient_)`?

contracts/deployables/warrant/WarrantBase.sol
```diff
@@ -0,0 +1,315 @@
...
+
+        // Check expiration based on time mode and token lock if relative time
+        if ($.relativeTime) {
+            if (IVotesERC20V1($.token).locked()) revert TokenLocked();
+            if (
+                block.timestamp >
+                IVotesERC20V1($.token).getUnlockTime() + $.expiration
+            ) revert Expired();
+        } else {
+            if (block.timestamp > $.expiration) revert Expired();
```
adamgall: suggestion: rename this error to `WarrantExpired`, to match the language of the `WarrantNotExpired` error.

contracts/deployables/warrant/WarrantBase.sol
```diff
@@ -0,0 +1,315 @@
...
+        // Delegate to implementation-specific logic
+        _executeWarrant(recipient_);
+
+        emit Executed(recipient_);
+    }
+
+    /**
+     * @inheritdoc IWarrantBase
+     */
+    function clawback(address recipient_) public virtual override onlyOwner {
```
adamgall, line 294: What if the owner calls `clawback()` twice? I assume the `IERC20($.token).safeTransfer(recipient_, $.tokenAmount);` line will fail? Is it safe to rely on that?

contracts/deployables/warrant/WarrantHedgeyV1.sol
```diff
@@ -0,0 +1,316 @@
...
+        internal
+        pure
+        returns (WarrantHedgeyStorage storage)
+    {
+        WarrantHedgeyStorage storage $;
+        assembly {
+            $.slot := WARRANT_HEDGEY_STORAGE_LOCATION
+        }
+        return $;
+    }
```
adamgall: Same comments here as in WarrantBase.sol regarding named return variable

contracts/deployables/warrant/WarrantHedgeyV1.sol
```diff
@@ -0,0 +1,316 @@
...
+
+    // ======================================================================
+    // CONSTRUCTOR & INITIALIZERS
+    // ======================================================================
+
+    constructor() WarrantBase() {}
```
adamgall: Change to:

```solidity
    constructor() {
        _disableInitializers();
    }
```

contracts/deployables/warrant/WarrantHedgeyV1.sol
```diff
@@ -0,0 +1,316 @@
...
+import {
+    IVotesERC20V1
+} from "../../interfaces/decent/deployables/IVotesERC20V1.sol";
+import {IVersion} from "../../interfaces/decent/deployables/IVersion.sol";
+import {IDeploymentBlock} from "../../interfaces/decent/IDeploymentBlock.sol";
+import {
+    DeploymentBlockInitializable
+} from "../../DeploymentBlockInitializable.sol";
+import {InitializerEventEmitter} from "../../InitializerEventEmitter.sol";
+import {IERC20} from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
```
adamgall, line 23: Is there any need to import and `using SafeERC20 for IERC20`?

contracts/deployables/warrant/WarrantHedgeyV1.sol
```diff
@@ -0,0 +1,316 @@
+// SPDX-License-Identifier: AGPL-3.0
+pragma solidity ^0.8.30;
+
+import {WarrantBase} from "./WarrantBase.sol";
```
adamgall: nitpick: pls put interface imports before WarrantBase

contracts/interfaces/decent/deployables/IWarrantHedgeyV1.sol
```diff
@@ -0,0 +1,128 @@
...
+        uint256 hedgeyPeriod;
+    }
+
+    // --- State-Changing Functions ---
+
+    /**
+     * @notice Initialize the warrant with Hedgey-specific parameters
+     * @param params_ Struct containing all initialization parameters
+     */
+    function initialize(InitParams calldata params_) external;
```
adamgall: this `initialize` function should be in a section called `// --- Initializer Functions ---` between the Events and the View Functions

contracts/deployables/warrant/WarrantHedgeyV1.sol
```diff
@@ -0,0 +1,316 @@
...
+
+        if (cliff_ > end) revert CliffExceedsEnd(cliff_, end);
+    }
+
+    /**
+     * @notice Implementation of warrant execution that creates a Hedgey vesting plan
+     * @dev Called by base contract after fee collection and validation
+     * @param recipient_ Address that will receive the vested tokens
+     */
+    function _executeWarrant(address recipient_) internal virtual override {
```
adamgall, line 213: This should be moved up into the `IWarrantHedgeyV1` section under a new subsection for `// --- Internal Functions ---`

adamgall (changes requested):

contracts/deployables/warrant/WarrantBase.sol
tbwebb22: Yeah `feeAmount` may not be the most accurate description, it is really the total amount the investor is paying to purchase their share of tokens.

`$.tokenPricePerFeeToken` kind of sounds backwards to me. Its a price per the `token`, not `feeToken` right? Maybe we need a more descriptive name for the token being bought than just `token` also

adamgall: I don't really have the answer here, you have way more context than I do.

Just from my own (naive) point of view, that line was kind of confusing to me.

I trust you to rename whatever needs to be renamed to make it clearer!

contracts/deployables/warrant/WarrantBase.sol
tbwebb22: removed it since we're not doing this check most places in the codebase

contracts/deployables/warrant/WarrantBase.sol
tbwebb22: Yes, that safeTransfer line will fail, and it's definitely safe to rely on that. We could add a check for it so we could throw a more descriptive error, but in my opinion its not necessary.

adamgall: when does the `tokenAmount` of `token`s get sent to this contract? What if this contract holds more `token`s than `tokenAmount`?

contracts/deployables/warrant/WarrantHedgeyV1.sol
tbwebb22: In my opinion its not necessary here, since there are only two ERC20s the contract is interacting with, both of which are passed in as initialization args, so we can assume they are trusted. Generally I'll use `SafeERC20` only when the contract can interact with any arbitrary ERC20.

adamgall: But you're using SafeERC20 in WarrantBase... or is that a different situation?

contracts/deployables/warrant/WarrantHedgeyV1.sol
tbwebb22: This is actually being overriden from `WarrantBase`, so it should go in a new `WarrantBase` section right?

adamgall: Yes absolutely, my mistake

contracts/deployables/warrant/WarrantBase.sol
tbwebb22: In this case I prefer keeping this check in place - since there is only a single incorrect state to check for. And these Warrant contracts could be in place for long periods of time - I'd prefer it to revert upon deployment rather than finding out something was initialized incorrectly after the long time period.

contracts/deployables/warrant/WarrantHedgeyV1.sol
tbwebb22: oh good call.. I'll remove it from WarrantBase

contracts/deployables/warrant/WarrantBase.sol
adamgall: hmm, ok that sounds fair.

Sounds like you're saying the first time someone would find out if a "relative time" config was deployed with a non-IVotesERC20 contract would be a long time after it was deployed?

if that's the case, then yeah i agree with you.

contracts/deployables/warrant/WarrantHedgeyV1.sol
adamgall: just as one final comment, i know that you HAVE to use SafeERC20 when doing `transferFrom` (i think it's that function) with USDT, for example.

i know there's some gas overhead to using it, but maybe just to be safe to make sure this this contract is compatible with USDT, we do use it?

contracts/deployables/warrant/WarrantBase.sol
tbwebb22: The expected flow is that the Countersign contract would have a list of TXs to execute, one of which would be deploying this Warrant contract, and a followup TX which transfers the correct token amount to this contract.

Currently, if this contract ends up with more `token` than `tokenAmount`, those tokens will be locked in the contract, with no way to rescue them. I remember discussing this with Eddie and John and we decided this was fine, but am open to counter arguments on it.

contracts/deployables/warrant/WarrantBase.sol
tbwebb22: yep, exactly

contracts/deployables/warrant/WarrantHedgeyV1.sol
tbwebb22: yeah that is valid. We probably do want this to be compatible with USDT. I'll add it back in both contracts

contracts/deployables/warrant/WarrantBase.sol
adamgall: if this was discussed and decided upon already, i'm not going to argue against it. just wanted to make sure it _has_ been discussed

contracts/deployables/warrant/WarrantHedgeyV1.sol
tbwebb22: I added `SafeERC20` back to `WarrantBase`. I just realized `WarrantHedgeyV1` doesn't actually have any ERC20 transfers, which is why it didn't import `SafeERC20` to begin with

adamgall: Do you wanna do `forceApprove` in WarrantHedgeyV1?

tbwebb22: Haven't used it before - here's the comment for `forceApprove`:

_Meant to be used with tokens that require the approval to be set to zero before setting it to a non-zero value, such as USDT_

I think its impossible for the allowance to be non-zero before we hit this approve line, do you agree?

adamgall: That's a good point. This contract is the only one that can "approve" itself, and _executeWarrant is only ever called one time. Think we're good here 👍

contracts/deployables/warrant/WarrantBase.sol
tbwebb22: Ok I renamed some variables to make things more explicit:

token => warrantToken
feeToken => paymentToken
tokenAmount => warrantTokenAmount
tokenPrice => warrantTokenPrice
feeReceiver => paymentReceiver

warrantTokenPrice units is explained in the WarrantBase storage struct:

/** @notice Price per warrant token in payment token units (18 decimal precision) */
        uint256 warrantTokenPrice;

adamgall (approved):

adamgall (changes requested):
Ah i almost forgot, can you please add this contract to the deployment scripts?

adamgall (approved):

PR #413, opened 2025-06-27
`[ENG-1238]` use AccessControlEnumerableUpgradeable...

twodam:
...so we can query list of whitelisted address for [this linear ticket](https://linear.app/decent-labs/issue/ENG-1070/stretch-add-transfer-whitelist-to-token-tab)

I'm not sure why `runSupportsInterfaceTests` failed with this new `IAccessControlEnumerable` interface. @adamgall

### DIFF
--- a/contracts/deployables/erc20/VotesERC20V1.sol
+++ b/contracts/deployables/erc20/VotesERC20V1.sol
@@ -16,8 +16,8 @@ import {
 } from "../../DeploymentBlockInitializable.sol";
 import {InitializerEventEmitter} from "../../InitializerEventEmitter.sol";
 import {
-    IAccessControl
-} from "@openzeppelin/contracts/access/IAccessControl.sol";
+    IAccessControlEnumerable
+} from "@openzeppelin/contracts/access/extensions/IAccessControlEnumerable.sol";
 import {ERC165} from "@openzeppelin/contracts/utils/introspection/ERC165.sol";
 import {
     ERC20Upgradeable
@@ -38,8 +38,8 @@ import {
     UUPSUpgradeable
 } from "@openzeppelin/contracts-upgradeable/proxy/utils/UUPSUpgradeable.sol";
 import {
-    AccessControlUpgradeable
-} from "@openzeppelin/contracts-upgradeable/access/AccessControlUpgradeable.sol";
+    AccessControlEnumerableUpgradeable
+} from "@openzeppelin/contracts-upgradeable/access/extensions/AccessControlEnumerableUpgradeable.sol";
 
 /**
  * @title VotesERC20V1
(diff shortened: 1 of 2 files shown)


adamgall:
@twodam i've fixed and pushed up a commit.

- added `IAccessControl` back into `supportsInterface` for `VotesERC20V1`
  - because `IAccessControlEnumerable` inherits from `IAccessControl`, and the way that solidity works, is that `type(IAccessControlEnumerable).interfaceId` _does not` include the functions from inherited interfaces (`IAccessControl`).
- fixed the tests to both check for `IAccessControl` interface support, and `IAccessControlEnumerable` support.
  - annoyingly, the typechain `IAccessControlEnumerable__factory` class _does_ include all of the functions from the inherited `IAccessControl__factory`, so in the helper function to generate `interfaceId` for a typechain class, we need to explicitly _remove_ the functions defined in `IAccessControl__factory` from `IAccessControlEnumerable__factory`.

adamgall (commented):
As a challenge to this PR:

Any reason why the offchain services displaying this data can't just use event logs to distill the current members of each role?

The `Enumerable` version of `AccessControl` is definitely necessary if another contract needs to get a list of all role members, but that's not part of our requirements.

https://docs.openzeppelin.com/contracts/5.x/api/access#AccessControl
> Contract module that allows children to implement role-based access control mechanisms. This is a lightweight version that doesn’t allow enumerating role members except through off-chain means by accessing the contract event logs. Some applications may benefit from on-chain enumerability, for those cases see [AccessControlEnumerable](https://docs.openzeppelin.com/contracts/5.x/api/access#AccessControlEnumerable).

johnqh:
@adamgall
Pro of using the events: Cheap gas
Cons: Additional dependencies

I would rather stick to one source of truth, which is the contract.

adamgall:
> @adamgall Pro of using the events: Cheap gas Cons: Additional dependencies
> 
> I would rather stick to one source of truth, which is the contract.

@johnqh  What additional dependencies are you referring to? A "simple" `.reduce()` in `viem` on those event logs gives us the set of current role members.

edit: but like whatever, i'm not gonna argue this one too hard. just want to make sure we're on the same page about what the tradeoffs are.

twodam:
I’ve managed to [parse the events](https://github.com/decentdao/decent-app/pull/3166/commits/7447c84eed1c1840d38501bf04666e5975b8a849#diff-5b2c1d6f7ed96d90d76baede56e74cdf5a16df604db0d666b4583ffbce5598c8R932-R956) to get the current whitelist, so the ticket no longer depends on this update. I’m fine with either putting this PR on hold or going ahead and merging it.

adamgall:
I would prefer to not merge this PR

johnqh:
Talked to @adamgall . I had misunderstanding that getting the events would take many round trips to the node, but actually the events are indexed on the node and can be retrieved with one viem call. So I agree with Adam

adamgall:
Closing this PR

PR #427, opened 2025-07-15
Kyc verify

tbwebb22:
This PR modifies the KYCVerifier contract to use signature verification. A backend server controlled by Decent will sign messages to attest to the caller's KYC status. The `operatingContract` - the contract checking the caller's KYC status is also signed in the message. In the current codebase these operating contracts could by either `KYCVerifierV1` or `CountersignV1`


contracts/services/kyc/KYCVerifierV1.sol
```diff
@@ -10,34 +10,56 @@ import {
...
+    // ======================================================================
+    // STATE VARIABLES
+    // ======================================================================
+
+    address private immutable _verifier;
+
+    bytes32 internal constant TYPEHASH =
+        keccak256(
+            "VerificationData(address operatingContract,address account)"
+        );
```
adamgall: I think the signatures should include a `nonce` and an `expiry` timestamp, to prevent replay attacks (track used nonces in contract state), and to prevent stale signatures from being used.

contracts/services/kyc/KYCVerifierV1.sol
```diff
@@ -10,34 +10,56 @@ import {
...
     DeploymentBlockNonInitializable,
-    ERC165
+    ERC165,
+    EIP712
 {
+    // ======================================================================
+    // STATE VARIABLES
+    // ======================================================================
+
+    address private immutable _verifier;
```
adamgall: Please fix the solhint warning on this line

contracts/services/kyc/KYCVerifierV1.sol
```diff
@@ -46,11 +68,30 @@ contract KYCVerifierV1 is
...
+        return
+            ECDSA.recover(
+                _hashTypedDataV4(
+                    keccak256(
+                        abi.encode(TYPEHASH, operatingContract_, account_)
+                    )
+                ),
+                signature_
+            ) == _verifier;
+    }
```
adamgall: maybe emit an event from this function?

contracts/interfaces/decent/services/IKYCVerifierV1.sol
```diff
@@ -13,7 +13,7 @@ pragma solidity ^0.8.30;
  * - Stateless view function for gas-efficient checks
```
adamgall: Is this still stateless?

contracts/deployables/countersign/CountersignV1.sol
```diff
@@ -341,7 +341,13 @@ contract CountersignV1 is
...
-        if (!IKYCVerifierV1($.kycVerifier).verify(msg.sender)) {
+        if (
+            !IKYCVerifierV1($.kycVerifier).verify(
+                address(this),
+                msg.sender,
+                verifyingSignature_
+            )
+        ) {
             revert InvalidKYCSignature();
         }
```
adamgall: just thinking here, what if the `verify` function on `KYCVerifierV1` threw the error if the signature isn't valid, instead of duplicating these errors in the calling contracts?

contracts/deployables/public-sale/PublicSaleV1.sol
```diff
@@ -106,11 +106,17 @@ contract PublicSaleV1 is
...
-            revert KYCVerificationFailed();
+        if (
+            !IKYCVerifierV1($.kycVerifier).verify(
+                address(this),
+                msg.sender,
+                verifyingSignature_
+            )
+        ) revert KYCVerificationFailed();
         _;
     }
```
adamgall: just thinking here, what if the `verify` function on `KYCVerifierV1` threw the error if the signature isn't valid, instead of duplicating these errors in the calling contracts?

contracts/services/kyc/KYCVerifierV1.sol
adamgall: and considering that this is not an upgradeable contract, and there's no way to change the `_verifier`, what is the remediation plan if the private key Decent is using on its server gets leaked? What would that affect and how would we fix it?

adamgall (changes requested):

contracts/services/kyc/KYCVerifierV1.sol
tbwebb22: In this context, how would a replay attack be used? Also, what does a stale signature mean in this context?

I do think the concept of an `expiry` timestamp could make sense here, but not really seeing the use-case for a nonce

contracts/services/kyc/KYCVerifierV1.sol
tbwebb22: But then this wouldn't be a view function

contracts/interfaces/decent/services/IKYCVerifierV1.sol
tbwebb22: The contract isn't, updated the comment there

contracts/deployables/countersign/CountersignV1.sol
tbwebb22: I'd prefer for `KYCVerifierV1` to just return true/false. Then the contract calling it have more flexibility. I can imagine situations where if a user isn't KYC verified, we want do some different logic, and not necessarily revert.

contracts/services/kyc/KYCVerifierV1.sol
adamgall: yeah good point

contracts/services/kyc/KYCVerifierV1.sol
tbwebb22: If the private key does get leaked, then the hacker can potentially generate signatures attesting to any address being KYC verified. They could potentially generate a bunch of these signatures for their own addresses and buy up a large portion of the public sale tokens. They would of course, still have to pay for them.

We could make `KYCVerifierV1` be `ownable` and the `owner` can update the `verifier` address

contracts/services/kyc/KYCVerifierV1.sol
adamgall: For example, without a protection for replay attacks, someone could use the same signature for multiple `increaseCommitment` calls, even if their KYC status has been revoked (in the offchain system) after the first call.

contracts/services/kyc/KYCVerifierV1.sol
adamgall: I assume Decent would be the owner of that contract? Maybe this does make sense, since it is _our address_ which is the verifier anyway.

contracts/deployables/countersign/CountersignV1.sol
adamgall: Yeah that's fair.

It's not even really a big deal, i just kind of paused when I saw the different errors being thrown from `CountersignV1` and `PublicSaleV1` for when `.verify()` returned false: `InvalidKYCSignature()`, vs `KYCVerificationFailed()`

contracts/services/kyc/KYCVerifierV1.sol
tbwebb22: ok yeah if revoking KYC status is something we need to support then I do agree with you

I'd like to run this one by John / Eddie to see if revoking KYC is a product requirement

contracts/services/kyc/KYCVerifierV1.sol
tbwebb22: Yeah, I could see an argument that the verifier private keys could be considered less secure since they will be managed by a server in some way. 

And the owner could be a more secure address, like a Decent multisig.

If this is true, it might make sense to separate them and make this contract ownable

contracts/deployables/countersign/CountersignV1.sol
tbwebb22: Ah yeah these should at least be the same error name - just updated so both errors are `KYCVerificationFailed`

contracts/services/kyc/KYCVerifierV1.sol
tbwebb22: @adamgall Eddie said we should support revoking KYC status. Do you agree that just having the signature expiration timestamp can satisfy this? We can just always keep the expiration timestamp very short into the future - like a couple minutes or so.

adamgall: Why wouldn't you want to include a nonce as well? Feels like that is best practice. Is it the gas overhead?

jigglyjams: I see the argument for adding both nonce and expiry. Very little overhead on the API, would need to fetch the latest nonce for that address and operatingContract

tbwebb22: Ok, I'll add nonces and expiration timestamps

contracts/services/kyc/KYCVerifierV1.sol
tbwebb22: Made `KYCVerifierV1` inherit from `Ownable2Step`, and added an `onlyOwner` `updateVerifier` function

contracts/services/kyc/KYCVerifierV1.sol
tbwebb22: this has been added, should be ready for a re-review

adamgall (changes requested):
now that there's a constructor in KYCVerifier, the deployment script will need to be updated.

also, i notice that we forgot to add the PublicSale contract into the deployment scripts. Can you please add that too?

contracts/services/kyc/KYCVerifierV1.sol
```diff
@@ -10,34 +10,66 @@ import {
...
+    EIP712,
+    Ownable2Step
 {
+    // ======================================================================
+    // STATE VARIABLES
+    // ======================================================================
+
+    address private _verifier;
+
+    mapping(address account => uint256 nonce) private _nonces;
```
adamgall, line 55: I know that _technically_ we don't need to, but what do you think about using namespaced storage for these variables?

The only reason I suggest it is because in `DeploymentBlockNonInitializable`, that's another contract that _technically_ doesn't need to use namespaced storage, but does anyway.

I'd like to keep it consistent, so either I think we should switch this contract to using namespaced storage, or rip it out of `DeploymentBlockNonInitializable`.

Thoughts?

tbwebb22: Personally I think we should remove the namespaced storage from `DeploymentBlockNonInitializable` as well then. It just adds complexity and there's no purpose for it in non-upgradeable contracts

adamgall: Ok, i'll agree with that!

I will work on that in a new PR.

AFAIK there aren't any other contracts that this applies to, but can you please double check that?

tbwebb22: just did pass through of all contracts - didn't see anywhere else this would apply to

tbwebb22:
> now that there's a constructor in KYCVerifier, the deployment script will need to be updated.
> 
> also, i notice that we forgot to add the PublicSale contract into the deployment scripts. Can you please add that too?

added to deploy scripts, should be ready for re-review

ignition/modules/services/ServicesModule.ts
```diff
@@ -1,8 +1,13 @@
 import { buildModule } from '@nomicfoundation/hardhat-ignition/modules';
+// eslint-disable-next-line import/no-extraneous-dependencies
```
adamgall: I'd prefer to just globally fix this by adding `"ignition/**/*.ts"` to line 53 of .eslintrc.json

adamgall (approved):

PR #428, opened 2025-07-22
refactor: Simplify storage handling in DeploymentBlockNonInitializable contract

adamgall:


contracts/DeploymentBlockNonInitializable.sol
```diff
@@ -34,40 +34,7 @@ abstract contract DeploymentBlockNonInitializable is IDeploymentBlock {
...
-        internal
-        pure
-        returns (DeploymentBlockNonInitializableStorage storage $)
-    {
-        // solhint-disable-next-line no-inline-assembly
-        assembly {
-            $.slot := DEPLOYMENT_BLOCK_NON_INITIALIZABLE_STORAGE_LOCATION
-        }
-    }
+    uint256 private _deploymentBlock;
```
tbwebb22: we should make this `immutable`

tbwebb22 (changes requested):

contracts/DeploymentBlockNonInitializable.sol
adamgall: oh yeah good point, fixed and pushed

tbwebb22 (approved):

PR #432, opened 2025-08-07
Public sale hedgey

tbwebb22:
This PR adds Hedgey integration to the Public Sale contract. A boolean passed in upon contract initialization indicates whether the Hedgey integration is used. If true - upon successful sale settlement, the Hedgey VotingTokenLockupPlans contract will be called to create a lockup plan for the buyer, and the buyer's corresponding sale token amount will be transferred to the Hedgey contract. If false upon a successful sale settlement - the buyer will simply be transferred their purchased sale token amount as before.

Added HedgeyLockupParams struct to initialization data:

`    struct HedgeyLockupParams {
        bool enabled;
        uint256 start;
        uint256 cliff;
        uint256 ratePercentage;
        uint256 period;
        address votingTokenLockupPlans;
    }
` 
The `ratePercentage` is different than the `rate` that is passed in to Hedgey, as it is different depending on the lockup amount. `ratePercentage` is represented as a percentage (divide the amount by 10**18) of the lockup amount that unlocks on each `period`.

So if ratePercentage is 0.001 * 10**18, and period is 10,000, then 0.1% of the lockup amount will unlock every 10,000 seconds.

`VotingTokenLockupPlans.sol` in the mocks/hedgey folder is a flattened version of the production Hedgey contract, but is functionally equivalent. This was done since Hedgey doesn't have a contracts npm package, but to still support integration testing.

### DIFF
--- a/contracts/deployables/public-sale/PublicSaleV1.sol
+++ b/contracts/deployables/public-sale/PublicSaleV1.sol
@@ -162,12 +162,18 @@ contract PublicSaleV1 is
 
         // if hedgey lockup is enabled, validate the params
         if (params_.hedgeyLockupParams.enabled) {
+            if (params_.commitmentToken == NATIVE_ASSET)
+                revert InvalidHedgeyNativeAsset();
+
+            uint256 minimumLockupAmount = (params_.minimumCommitment *
+                PRECISION) / params_.saleTokenPrice;
+
             _validateHedgeyParams(
-                (params_.minimumCommitment * PRECISION) /
-                    params_.saleTokenPrice,
+                minimumLockupAmount,
                 params_.hedgeyLockupParams.start,
                 params_.hedgeyLockupParams.cliff,
-                params_.hedgeyLockupParams.rate,
+                (minimumLockupAmount *
+                    params_.hedgeyLockupParams.ratePercentage) / PRECISION,
                 params_.hedgeyLockupParams.period
             );
         }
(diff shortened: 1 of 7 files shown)


contracts/deployables/public-sale/PublicSaleV1.sol
```diff
@@ -599,6 +728,40 @@ contract PublicSaleV1 is
...
+     * @param amount_ Total amount to vest
+     * @param rate_ Amount vested per period
+     * @param period_ Duration of each vesting period
+     * @custom:throws InvalidAmount if amount is zero
+     * @custom:throws InvalidRate if rate is zero
+     * @custom:throws RateExceedsAmount if rate is greater than amount
+     * @custom:throws InvalidPeriod if period is zero
+     * @custom:throws CliffExceedsEnd if cliff time exceeds vesting end time
+     */
+    function _validateHedgeyParams(
```
twodam, line 745: maybe we should validate `votingTokenLockupPlans` contract too

test/unit/deployables/erc20/VotesERC20StakedV1.test.ts
```diff
@@ -36,7 +36,10 @@ async function runExecuteTxAndCheckBalanceDeltasTests(
       if (t.token === 'native') {
         return ethers.provider.getBalance(t.addressToCheck);
       } else {
-        const token = await ethers.getContractAt('IERC20', t.token);
+        const token = await ethers.getContractAt(
+          '@openzeppelin/contracts/token/ERC20/IERC20.sol:IERC20',
```
twodam, line 40: What’s the reason for this change? I don’t see any interface update on the `VotesERC20StakedV1` contract, so I’m wondering why this needs to be updated here in the test file.

contracts/deployables/public-sale/PublicSaleV1.sol
```diff
@@ -474,14 +578,39 @@ contract PublicSaleV1 is
...
+                // approve hedgey lockup plan to transfer the sale token
+                IERC20($.saleToken).approve(
+                    $.hedgeyLockupParams.votingTokenLockupPlans,
+                    saleTokenAmount
+                );
+
+                // create hedgey lockup plan for the caller
+                IVotingTokenLockupPlans(
+                    $.hedgeyLockupParams.votingTokenLockupPlans
+                ).createPlan(
```
twodam, line 596: I noticed there’s a `getLockupPlanCreatedEvent` function in the test file that retrieves the planId from the Hedgey event. Could it be that we might also need the planId on the frontend later? If so, maybe we could update the `SuccessfulSaleSettled` event to include the `planId` returned from the `createPlan` function call.

hardhat.config.ts
```diff
@@ -15,6 +15,18 @@ const config: HardhatUserConfig = {
           },
         },
       },
+      {
+        // 0.8.19 compiler is only used for the VotingTokenLockupPlans contract,
```
twodam: To make the compiler version explicit for a single contract only, we could use [solidity.overrides](https://hardhat.org/hardhat-runner/docs/advanced/multiple-solidity-versions):

```
solidity: {
  compilers: [...],
  overrides: {
      "contracts/mocks/hedgey/VotingTokenLockupPlans.sol": {
        version: '0.8.19',
        settings: {
          optimizer: {
            enabled: true,
            runs: 200,
          },
          viaIR: true,
        },
      }
    }
```

twodam (changes requested):

test/unit/deployables/public-sale/PublicSaleV1.test.ts
```diff
@@ -182,6 +186,37 @@ async function expectSaleState(sale: PublicSaleV1, expectedState: SaleState): Pr
...
+      return parsed?.name === 'PlanCreated';
+    } catch {
+      return false;
+    }
+  });
+
+  if (planCreatedEvent) {
+    const parsedEvent = votingTokenLockupPlans.interface.parseLog(planCreatedEvent);
+    return {
+      planId: parsedEvent?.args?.[0],
```
twodam, line 205: only `planId` is referenced in tests, maybe remove other fields

contracts/deployables/public-sale/PublicSaleV1.sol
tbwebb22: how so - Just checking that its not address zero?

test/unit/deployables/erc20/VotesERC20StakedV1.test.ts
tbwebb22: so without this change, you will get this error:
image

Since the mock VotingTokenLockupPlans is a flattened version of their file, it declares IERC20. So now we have two artifacts that are IERC20. So we need to specify here which one we are using.

contracts/deployables/public-sale/PublicSaleV1.sol
tbwebb22: Yes, this is actually added in PR #433.

hardhat.config.ts
tbwebb22: Oh yeah I like this. Pushed this change up

test/unit/deployables/public-sale/PublicSaleV1.test.ts
tbwebb22: done

contracts/deployables/public-sale/PublicSaleV1.sol
twodam: > how so - Just checking that its not address zero?

yes, as an extra check when deploying&initializing the contract?

twodam (approved):
Approve as some comments has been addressed.

contracts/deployables/public-sale/PublicSaleV1.sol
tbwebb22: Yeah so a month or two ago, Adam and I agreed to remove the address(0) checks in the codebase since there's a near infinite number of incorrect addresses that could be configured, and address(0) is just one of them.

PR #433, opened 2025-08-08
Remove ownable public sale

tbwebb22:
This PR: 
- Removes `Ownable2StepUpgradeable` inheritance from `PublicSaleV1`
- Replaces references to `owner` with `seller`
- Adds reference to `buyer` in functions and events.
- Creates a new event `SuccessfulSaleBuyerSettledHedgey` that emits the created lockup plan ID

### DIFF
--- a/contracts/deployables/public-sale/PublicSaleV1.sol
+++ b/contracts/deployables/public-sale/PublicSaleV1.sol
@@ -561,7 +561,7 @@ contract PublicSaleV1 is
     /**
      * @inheritdoc IPublicSaleV1
      */
-    function settle(address recipient_) public virtual override {
+    function buyerSettle(address recipient_) public virtual override {
         PublicSaleStorage storage $ = _getPublicSaleStorage();
 
         if ($.settled[msg.sender]) revert AlreadySettled();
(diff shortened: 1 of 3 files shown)


twodam:
Should this PR be based on #432? My usual flow for dependent PRs is to branch the following PR from the previous PR’s branch. When merging, we merge the earlier PR first, and the later one will then automatically rebase onto `develop`.

contracts/deployables/public-sale/PublicSaleV1.sol
```diff
@@ -603,12 +598,23 @@ contract PublicSaleV1 is
                             $.hedgeyLockupParams.ratePercentage) / PRECISION,
                         $.hedgeyLockupParams.period
                     );
+
+                emit SuccessfulSaleBuyerSettledHedgey(
```
twodam, line 602: would it be better to emit a single `SuccessfulSaleBuyerSettled` event with an extra `hedgeyLockupPlanId` field, which could be 0 if no lockup plan is involved?

twodam (approved):

contracts/deployables/public-sale/PublicSaleV1.sol
tbwebb22: I have a slight preference towards keeping it as, and only emitting what is relevant to the sale type. But if you can give a strong argument to change it to a single event, I'm open to it.

tbwebb22:
> Should this PR be based on #432? My usual flow for dependent PRs is to branch the following PR from the previous PR’s branch. When merging, we merge the earlier PR first, and the later one will then automatically rebase onto `develop`.

Good call, looks like the base has already been updated to `public-sale-hedgey`

PR #434, opened 2025-08-13
add sale token protocol fee

tbwebb22:
This PR adds a configurable protocol fee percentage paid by the seller in the fee token. Upon initialization, the contract transfers from the seller and escrows the maximum amount of sale tokens that may be needed. This is the maximum amount of sale tokens that could be sold plus the sale token protocol fee. At the end of the sale, if not all of the sale tokens are sold, the seller gets refunded the leftover amount of sale tokens when the `sellerSettle` function is called.


cbajandas (approved):

dacebt (approved):
I probably need to dive more into PRD and compare the actual logic flow but code looks good

PR #445, opened 2025-10-18
`[ENG-1540]` Update Sablier in Role Utility Management Contract to V2

dacebt:


twodam (approved):
it looks like only some type changes are applied for Sablier V2, no breaking logic introduced.

